PR #2, opened 2017-08-18
remove beta paragraph

gidsg:
I found this paragraph confusing and a bit inaccurate as we don't use the terminology 'live' anywhere else. Beta is our equivalent of a live environment right now.

If you don't think it should be removed might be possible to re-write it to something more useful?

CC @J-Lambo @ellayanor


J-Lambo:
Hi @gidsg - sorry it's taken me a while to review this! Good spot and good suggestion. I'm happy to remove.

PR #3, opened 2017-09-18
Update API reference

frankieroberto:
Entries endpoint seems to use `start` and `limit` instead of `page-index` and `page-size`.

### DIFF
--- a/source/documentation/using_the_api/api_reference.md
+++ b/source/documentation/using_the_api/api_reference.md
@@ -205,7 +205,7 @@ Path: GET /entries
 
 Get all entries from the register. For example, all updates there have ever been to the country register.
 
-*Note: Results from this API call are paginated. This call will return the first 100 entries from the first page of the register. Use `page-size` to define the number of entries you want and `page-index` to define the pages you want. The maximum `page-size` is 5000.*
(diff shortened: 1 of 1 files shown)


karlbaker02:
For `/records`, `page-index` and `page-size` are valid query parameters; for `/entries`, we use `start` and `limit` to mean the same thing. We should be consistent in what we're doing. Thoughts @michaelabenyohai @gidsg

frankieroberto:
@karlbaker02 as I understand it, `page-size` and `limit` are the same thing, but `page-index` and `from` are different ways of doing pagination (the former is referencing a page-number, e.g. page 2 of 10-per-page returns 11-20, whereas from is the item to start from).

For what it's worth, I'd suggest that that `from`and `limit` are a better way of doing pagination, as they're more flexible, and solve the problem of seeing duplicate/missing entries if an entry was added or deleted between requests (although this mostly only happens if you're sorting in descending timestamp order).

frankieroberto:
@karlbaker02 can this be merged whilst you consider the longer-term possibilities of rationalising the query param names?  Currently the documentation is wrong, and this fixes that. (Just caught me out again).

frankieroberto:
@michaelabenyohai @bahmady thanks both.

PR #9, opened 2017-11-29
Changes to api_reference.md subheadings (and ToC)

bahmady:
User research feedback suggested that when looking at the ToC, users felt the headings such as 'View a specific record' and 'View all entires' weren't technical enough and didn't make enough sense by themselves in the context of REST APIs. They felt that the information only made sense when they looked at the contents of each section. 

The changes in this commit change the subheadings (and therefore the ToC) to more technical names, to aid searching. This will be fed into the next round of UR in week 9 of Q3 to ascertain whether in this new form, this is easier for users to follow.

### DIFF
--- a/source/documentation/using_the_api/api_reference.md
+++ b/source/documentation/using_the_api/api_reference.md
@@ -1,8 +1,6 @@
 ## <a name="apireference"></a>API reference
 
-### Find out more about a register
-
-Path: GET /register
+### GET /register
 
 View information about a register, including:  
 
@@ -43,9 +41,7 @@ Example response:
 }
 ```
 
-### Get all records
-
-Path: GET /records
+### GET /records
 
 Get all records from the register. For example, all of the countries from the country register.
 
@@ -106,9 +102,7 @@ Example response:
 }
 ```
 
-### View a specific record
-
-Path: GET /record/{field-value}
+### GET /record/{field-value}
 
 Find a specific record within a register. For example, the record for Vatican City in the country register.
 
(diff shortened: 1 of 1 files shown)


source/documentation/using_the_api/api_reference.md
```diff
@@ -1,8 +1,6 @@
 ## <a name="apireference"></a>API reference
 
-### Find out more about a register
-
-Path: GET /register
+### GET /register
```
arnau: Is it possible to make these headers monospaced? E.g. `GET /register`. Helps when discriminating between prose and code.

source/documentation/using_the_api/api_reference.md
```diff
@@ -374,21 +358,17 @@ You can download the new entries in 2 ways:
...
 
 You can find the latest entry number by looking at the register information and comparing the most recent entry number with your own copy.
 
 Example request: https://country.register.gov.uk/register
 
-### Download a full new copy of the register
+### <a name="downloadreg"></a>GET /download-register
```
arnau: Why some headers have an anchor and others don't?

bahmady: I left those in because they were in the original; I assumed those were linked to elsewhere (which seems to be the case) but there's no reason not to add anchors to all headers, anyway. The only reason I have refrained from opening a PR to add anchors to do that yet is because we're still changing them.

source/documentation/using_the_api/api_reference.md
bahmady: Yes, I can do that, but they won't appear monospaced in the ToC.

PR #10, opened 2017-11-30
Updated email address in rate_limits.md

bahmady:
Updated email address to registers@digital.cabinet-office.gov.uk


gidsg:
I know this is probably out of scope for this PR, but it makes no sense that the email address is in `rate_limits.md` should be a separate include?!

bahmady:
Yes, I agree. I'll open a separate PR with this when we have a better idea of how to restructure the API docs as a whole (alongside many other PRs).

PR #11, opened 2017-12-01
Rearrangement of information in api_reference.md

bahmady:
Moved information about `GET /register` in `Getting updates` section to first section on `GET /register` (partial duplication removed but the first section now has slightly more information)

Removed duplicated information about `GET /download-register` under original `Getting updates` section

Reorganised structure of information and headings in `Getting updates` section to start to move towards a tutorial-like format

### DIFF
--- a/source/documentation/using_the_api/api_reference.md
+++ b/source/documentation/using_the_api/api_reference.md
@@ -8,7 +8,7 @@ View information about a register, including:
 * when the register was last updated
 * how many entries and records the register contains
 
-You can find the latest entry number by looking at the register information and comparing the most recent entry number with your own copy.
+You can check the entry number you have locally is up to date with the register by comparing the value of `total-entries` for each. 
 
 Example URL: `https://local-authority-eng.register.gov.uk/register`
 
@@ -340,24 +340,7 @@ Example response:
 }
 ```
 
-## Getting updates
-
-You can download the new entries in 2 ways:  
-
-* download a copy of the full register
-* download individual updated entries for the records you are using
-
-### <a name="downloadreg"></a>`GET /download-register`
-
-Download the full contents of the register in a ZIP file.
-
-*Note: This will download every entry and item as an individual JSON file. If you only want to download records, use `GET /records`.*
-
-Example URL: `https://local-authority-eng.register.gov.uk/download-register`
-
-Example request: `curl -o localauthorityeng.zip --request GET --url https://local-authority-eng.register.gov.uk/download-register`
-
-### `GET /record/{field-value}/entries`
+### <a name="entries">`GET /record/{field-value}/entries`
 
 Get all entries for a single record.
 
@@ -391,3 +374,24 @@ Example response:
 ```
 
 You can then download the latest item, for example entry 204 in the above example. Follow the [guidance for downloading items](#items).
+
+### <a name="download">`GET /download-register`
+
+Download the full contents of the register in a ZIP file.
+
+*Note: This will download every entry and item as an individual JSON file. If you only want to download records, use `GET /records`.*
+
+Example URL: `https://local-authority-eng.register.gov.uk/download-register`
+
+Example request: `curl -o localauthorityeng.zip --request GET --url https://local-authority-eng.register.gov.uk/download-register`
+
+## Getting updates
+
+You can download new entries in 2 ways:  
+
+* [download a copy of the full register](#download)
+* [GET individual new entries for the records you are using](#entries)
+
+
+
+
--- a/source/documentation/using_the_api/api_reference.md
+++ b/source/documentation/using_the_api/api_reference.md
@@ -340,7 +340,7 @@ Example response:
 }
 ```
 
-### <a name="entries">`GET /record/{field-value}/entries`
+### <a name="record-entries">`GET /record/{field-value}/entries`
 
 Get all entries for a single record.
 
--- a/source/documentation/using_the_api/api_reference.md
+++ b/source/documentation/using_the_api/api_reference.md
@@ -8,6 +8,8 @@ View information about a register, including:
 * when the register was last updated
 * how many entries and records the register contains
 
+You can find the latest entry number by looking at the register information and comparing the most recent entry number with your own copy.
+
 Example URL: `https://local-authority-eng.register.gov.uk/register`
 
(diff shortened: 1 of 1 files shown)


bahmady:
@arnau @michaelabenyohai As you both picked up on the problems with the 'Getting updates' section I've made a start on reorganising the information

I'd like to use this PR as a place to openly discuss how best to do this since it seems to make more sense. I don't anticipate that the changes I've made will automatically be the best way to structure the information, so I'd love to get your feedback on:

* whether the broad changes I've made to the way the information is structured make sense, or not
* if anything vital is now missing, or is more ambiguous than before
* what needs to be further elaborated / expanded on

source/documentation/using_the_api/api_reference.md
```diff
@@ -338,32 +340,22 @@ Example response:
...
-
-Example request: `curl -o localauthorityeng.zip --request GET --url https://local-authority-eng.register.gov.uk/download-register`
-
 ## Getting updates
 
 You can download the new entries in 2 ways:  
 
-* download another copy of the full register
+* download a copy of the full register
 * download individual updated entries for the records you are using
```
arnau: If I'm being picky "updated entries" is abuse of language given that entries are immutable. Records on the other hand are recomputed for any change in a Register. @michaelabenyohai am I correct?

"Could it be changed to something on the lines of "download new entries for the records you are interested in". (I don't know if I like the "you are using/you are interested in" but besides the point I want to get across.

arnau:
@michaelabenyohai Another option I was mentioning to @bahmady is to get rid of the "Getting updates" section for the time being and move the endpoint definitions out to their resource groups (i.e. download alone, `record/{x}/entries` below `/record/{x}`.

source/documentation/using_the_api/api_reference.md
bahmady: My understanding of the language is the same. "Download new entries for the records you are interested in" sounds like quite a mouthful, but it gets the point across better.

To my mind, the problem is if you don't say something like "...you are interested in" then you're left with something like "...of interest" or "...in question" and out of context, those doesn't make as much sense. So "...you are interested in" is a bit of a mouthful, but it might be the least worst option for now.

bahmady:
@Maidment64 This isn't currently in Trello. We could spin out a card from https://trello.com/c/abi4GhdJ/1215-2-iterate-api-documentation-based-on-recent-testing

source/documentation/using_the_api/api_reference.md
```diff
@@ -8,6 +8,8 @@ View information about a register, including:
 * when the register was last updated
 * how many entries and records the register contains
 
+You can find the latest entry number by looking at the register information and comparing the most recent entry number with your own copy.
+
```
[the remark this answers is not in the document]
michaelabenyohai: Agree that this could be reworded slightly. Getting `total-entries` from this endpoint gives you the latest *entry number*. Comparing this number to the latest entry number you have locally tells you whether you are up to date with the register.

source/documentation/using_the_api/api_reference.md
michaelabenyohai: Agree that "updated entries" is not correct and we should be talking about "new entries" instead. As you say records change (as a function of the new entries in the register) - records are the things that are mutable, not entries.

michaelabenyohai: Also - what do we think about using the word "download" in the context of API usage? I'm possibly being a bit picky here and maybe it's more relevant to a different PR. Should we use "get" instead? (Unless we are explicitly talking about the `/download-register` link, where it does make sense as it downloads files.

arnau: What do you mean by "files"?

michaelabenyohai: It downloads a `.zip` file

arnau: So, being picky, it lets you get the whole register as `application/zip` (and once uncompressed you get a set of JSON files)?

I'm not particularly bothered by the word "download" but I'm fine keeping everything as "get" when you ask for a resource and "download" when you ask for a compressed dump.

michaelabenyohai:
I wonder, similarly whether it's worth moving all endpoints into API reference sections and keeping the Getting Updates section in terms of things that exist in the API reference section. I'm not sure it makes sense to only split `/record/{field-value}/entries` into the Getting Updates section when it doesn't exist in in the API reference section. I don't see it as specific to getting updates.

bahmady:
I've tried to pull in all of the different points of this conversation into https://github.com/alphagov/registers-tech-docs/pull/11/commits/1bb19c034f9229b83873e2d5a4f9b4f6bb8ca110

source/documentation/using_the_api/api_reference.md
```diff
@@ -338,34 +340,7 @@ Example response:
...
-You can find the latest entry number by looking at the register information and comparing the most recent entry number with your own copy.
-
-Example request: `https://local-authority-eng.register.gov.uk/register`
-
-### <a name="downloadreg"></a>`GET /download-register`
-
-See [endpoint for downloading register](#downloadreg).
-
-### `GET /record/{field-value}/entries`
+### <a name="entries">`GET /record/{field-value}/entries`
```
arnau: I'd probably use `<a name="record-entries">` to distinguish it from the `/entries` endpoint.

bahmady:
Is the order of these headers OK? Currently it's:

`GET /register`
`GET /records`
`GET /record/{field-value}`
`GET /records/{field-name}/{field-value}`
`GET /entries`
`GET /entry/{entry-number}`
`GET /item/{item-hash}`
`GET /record/{field-value}/entries`
`GET /download-register`
`Getting updates`

arnau:
Mmm, I would keep the `GET /record/{field-value}/entries` close to the other Record endpoints.

bahmady:
I was thinking this order(?):

`GET /register`
`GET /records`
`GET /records/{field-name}/{field-value}`
`GET /record/{field-value}`
`GET /record/{field-value}/entries`
`GET /entries`
`GET /entry/{entry-number}`
`GET /item/{item-hash}`
`GET /download-register`
`Getting updates`

arnau:
Looks consistent to me 👍

bahmady:
This PR ended up covering a lot more than I originally envisaged, so I've renamed it accordingly.

PR #12, opened 2017-12-06
remove trailing slash

gidsg:
CC @bahmady


bahmady:
Should the trailing slash in:

`https://local-authority-eng.register.gov.uk/`

not go as well? 

I think the reason this got missed is because it's in quick_start.md which is separate to api_reference.md where the rest of the docs live.

gidsg:
@benh-gds Nope that is the hostname fully qualified which is correct. But the path fragment shouldn't have a trailing slash.

bahmady:
OK, cool. Approving

PR #25, opened 2018-03-06
Renames `troubleshooting.md` to `support.md`

bahmady:
Based on user testing, we know users aren't looking for general support information. This was confirmed in the testing environment. This PR is to address that, and to make `troubleshooting.md` focused around support of the documentation.


source/documentation/support/support.md
```diff
@@ -0,0 +1,34 @@
+## Troubleshooting
```
arnau: If you changed the name of the file to "support.md", shouldn't the document title be "Support"?

Also, why is the first header a h2 instead of a h1? Is it because multiple files build up to a single html?

source/documentation/support/support.md
```diff
@@ -0,0 +1,34 @@
...
+
+If you cannot find a specific record, check you have the correct field value. You must use the exact field value to get a match from the register. For example, in the `local-authority-eng` register, you must capitalise the field value: 
+
+Incorrect request: https://local-authority-eng.register.gov.uk/record/bir
+
+Correct request: https://local-authority-eng.register.gov.uk/record/BIR
+
+### Adding additional data 
+
+If you want to combine register data with data from other sources, you may find it useful to keep the data from each source in separate files and combine them at a later stage. For example, you could keep separate files for data about countries and for data about territories.
```
arnau: I'm not following what do you mean here.

bahmady: This is a holdover from before my time. It seems faintly useless to me but I assumed it was important.

source/documentation/support/support.md
bahmady: Yes, you're right re. the first point. 

On the second point, yes, it is indeed.

source/documentation/support/support.md
bahmady: @arnau I'll open a separate PR to address this since this one's approved.

PR #29, opened 2018-03-16
Remove duplicate API reference sections

michaelabenyohai:
We had accidentally documented the following endpoints twice so I
have removed the duplicates.

- /record/{field-value}
- /record/{field-value}/entries


bahmady:
Thanks @michaelabenyohai — sorry for this oversight.

bahmady:
I'm going to merge and deploy this since I assume @michaelabenyohai has no objections and it doesn't look good to have duplicates.

PR #33, opened 2018-03-28
Change quick start guide

subatoi:
This in effect simplifies and expands explanations of base URLs and endpoints, beyond simply extending the quick start guide

### DIFF
--- a/source/documentation/getting_started/quick_start.md
+++ b/source/documentation/getting_started/quick_start.md
@@ -14,21 +14,21 @@ You can find a full list of register base URLs on our [Registers collection page
 | `country` | `https://country.register.gov.uk/` |
 | `allergen`  | `https://allergen.register.gov.uk/` |
 
-**2. [Find the endpoints you want to use](#apireference).**
+**2. Find the endpoints you want to use.**
 
 For example, `GET record/BIR` is the endpoint for Birmingham City Council in the `local-authority-eng` register.
 
 | Endpoint | Description |
 |----------|----------|
-| `GET /register`     | View information about a register. |
-| `GET /records` | Get all records from a register. |
-| `GET /record/{field-value}`  | Find a specific record within a register. |
-| `GET /record/{field-value}/entries`  | Get all entries for a single record. |
-| `GET /records/{field-name}/{field-value}`  | Find all records that share a field-value for a particular field. |
-| `GET /entries`  | Get all entries from a register. |
-| `GET /entry/{entry-number}`     | Find a specific entry from a register. |
-| `GET /item/{item-hash}` | Find a specific item within a register. |
-| `GET /download-register`  | Download the full contents of a register in a ZIP file. |
+| [`GET /register`](#get-register)     | View information about a register. |
+| [`GET /records`](#get-records) | Get all records from a register. |
+| [`GET /record/{field-value}`](#get-record-field-value)  | Find a specific record within a register. |
+| [`GET /record/{field-value}/entries`](#get-record-field-value-entries)  | Get all entries for a single record. |
+| [`GET /records/{field-name}/{field-value}`](#get-records-field-name-field-value) | Find all records that share a field-value for a particular field. |
+| [`GET /entries`](#get-entries)  | Get all entries from a register. |
+| [`GET /entry/{entry-number}`](#get-entry-entry-number)     | Find a specific entry from a register. |
+| [`GET /item/{item-hash}`](#get-item-item-hash) | Find a specific item within a register. |
+| [`GET /download-register`](#get-download-register)  | Download the full contents of a register in a ZIP file. |
 
 **3. Add `.json`, `.yaml`, `.csv`, `.tsv`, or `.ttl` to the end of the request URL (before any query parameters) to choose the format of the response, or amend headers to `accept application/json` or equivalent.**
 
--- a/source/documentation/getting_started/quick_start.md
+++ b/source/documentation/getting_started/quick_start.md
@@ -1,10 +1,37 @@
 ## Quick start guide
 
-Registers in beta are reliable and are ready for use in products and services.
+Registers which are in beta are reliable and ready for use in products and services. Each register has an open, RESTful API which you can use to access the data in the register. 
 
-1. [Find the base URL for the register(s) you want to use](https://registers.cloudapps.digital/registers). For example, for the `local-authority-eng` register: `https://local-authority-eng.register.gov.uk/`.
-2. [Find the endpoints you want](#apireference). For example, `/record/BIR` in `https://local-authority-eng.register.gov.uk/record/BIR`.
-3. Add `.json`, `.yaml`, `.csv`, `.tsv`, or `.ttl` to the end of the request URL (before any query parameters) to choose the format of the response, or amend headers to `accept application/json` or equivalent.
-4. Parse register data and use it in your product or service.
+You can follow this guide to quickly get started with GOV.UK Registers APIs. 
+
+**1. Find the base URL for the register(s) you want to use.** 
+
+You can find a full list of register base URLs on our [Registers collection page] (https://registers.cloudapps.digital/registers), but the following are some examples:
+
+| Register name | Base URL |
+|----------|----------|
+| `local-authority-eng`     | `https://local-authority-eng.register.gov.uk/`|
+| `country` | `https://country.register.gov.uk/` |
+| `allergen`  | `https://allergen.register.gov.uk/` |
+
+**2. [Find the endpoints you want to use](#apireference).**
+
+For example, `GET record/BIR` is the endpoint for Birmingham City Council in the `local-authority-eng` register.
+
+| Endpoint | Description |
+|----------|----------|
+| `GET /register`     | View information about a register. |
+| `GET /records` | Get all records from a register. |
+| `GET /record/{field-value}`  | Find a specific record within a register. |
+| `GET /record/{field-value}/entries`  | Get all entries for a single record. |
+| `GET /records/{field-name}/{field-value}`  | Find all records that share a field-value for a particular field. |
+| `GET /entries`  | Get all entries from a register. |
+| `GET /entry/{entry-number}`     | Find a specific entry from a register. |
+| `GET /item/{item-hash}` | Find a specific item within a register. |
+| `GET /download-register`  | Download the full contents of a register in a ZIP file. |
+
+**3. Add `.json`, `.yaml`, `.csv`, `.tsv`, or `.ttl` to the end of the request URL (before any query parameters) to choose the format of the response, or amend headers to `accept application/json` or equivalent.**
+
+**4. Parse register data and use it in your product or service.**
 
 Most of the API functions return paginated results. Follow the specific guidance for each endpoint to get the page(s) of entries, items, and records you need. Use `page-size` to define the number of results you want and `page-index` to define the pages you want. The maximum `page-size` is 5000.
--- a/source/documentation/getting_started/quick_start.md
+++ b/source/documentation/getting_started/quick_start.md
@@ -2,8 +2,6 @@
 
 Each register has an open, RESTful API which you can use to access the data in the register. 
 
-You can follow this guide to quickly get started with GOV.UK Registers APIs. 
-
 **1. Find the base URL for the register(s) you want to use.** 
 
 You can find the base URL for each register from the API inspector. For example, for the `local-authority-eng` register the API inspector is located at `https://local-authority-eng.register.gov.uk/`. 
--- a/source/documentation/getting_started/quick_start.md
+++ b/source/documentation/getting_started/quick_start.md
@@ -1,16 +1,16 @@
 ## Quick start guide
 
-Registers which are in beta are reliable and ready for use in products and services. Each register has an open, RESTful API which you can use to access the data in the register. 
+Each register has an open, RESTful API which you can use to access the data in the register. 
 
 You can follow this guide to quickly get started with GOV.UK Registers APIs. 
 
 **1. Find the base URL for the register(s) you want to use.** 
 
-You can find a full list of register base URLs [here](https://registers.cloudapps.digital/registers).
+You can find the base URL for each register from the API inspector. For example, for the `local-authority-eng` register the API inspector is located at `https://local-authority-eng.register.gov.uk/`. 
 
-The following are some examples of base URLs:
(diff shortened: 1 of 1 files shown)


source/documentation/getting_started/quick_start.md
```diff
@@ -1,10 +1,37 @@
...
 
-1. [Find the base URL for the register(s) you want to use](https://registers.cloudapps.digital/registers). For example, for the `local-authority-eng` register: `https://local-authority-eng.register.gov.uk/`.
-2. [Find the endpoints you want](#apireference). For example, `/record/BIR` in `https://local-authority-eng.register.gov.uk/record/BIR`.
-3. Add `.json`, `.yaml`, `.csv`, `.tsv`, or `.ttl` to the end of the request URL (before any query parameters) to choose the format of the response, or amend headers to `accept application/json` or equivalent.
-4. Parse register data and use it in your product or service.
+You can follow this guide to quickly get started with GOV.UK Registers APIs. 
+
+**1. Find the base URL for the register(s) you want to use.** 
+
+You can find a full list of register base URLs on our [Registers collection page] (https://registers.cloudapps.digital/registers), but the following are some examples:
```
danmitchell-: Not sure if `Registers collection page` is the correct term now. Might be worth checking with Connor - https://registers-trial.service.gov.uk/registers

source/documentation/getting_started/quick_start.md
```diff
@@ -1,10 +1,37 @@
...
+
+**2. Find the endpoints you want to use.**
+
+For example, `GET record/BIR` is the endpoint for Birmingham City Council in the `local-authority-eng` register.
+
+| Endpoint | Description |
+|----------|----------|
+| [`GET /register`](#get-register)     | View information about a register. |
+| [`GET /records`](#get-records) | Get all records from a register. |
+| [`GET /record/{field-value}`](#get-record-field-value)  | Find a specific record within a register. |
```
danmitchell-: Could `/field-value` be better described as `/key` i.e. not to confuse with `fields` 🤔

danmitchell-:
This is a fantastic improvement. Will let the powers that be have a final say ;-)

source/documentation/getting_started/quick_start.md
```diff
@@ -1,10 +1,37 @@
...
+| [`GET /record/{field-value}/entries`](#get-record-field-value-entries)  | Get all entries for a single record. |
+| [`GET /records/{field-name}/{field-value}`](#get-records-field-name-field-value) | Find all records that share a field-value for a particular field. |
```
bahmady: Probably need to find a way to avoid saying "that share a field value" so @danmitchell-'s idea of renaming it "key" could be useful here.

source/documentation/getting_started/quick_start.md
```diff
@@ -1,10 +1,37 @@
 ## Quick start guide
 
-Registers in beta are reliable and are ready for use in products and services.
+Registers which are in beta are reliable and ready for use in products and services. Each register has an open, RESTful API which you can use to access the data in the register.
```
bahmady: Should we be describing this as a "RESTful" or a "REST" API? Information I've read online and conversations I've had are conflicting.

source/documentation/getting_started/quick_start.md
```diff
@@ -1,10 +1,37 @@
...
+|----------|----------|
+| `local-authority-eng`     | `https://local-authority-eng.register.gov.uk/`|
+| `country` | `https://country.register.gov.uk/` |
+| `allergen`  | `https://allergen.register.gov.uk/` |
+
+**2. Find the endpoints you want to use.**
+
+For example, `GET record/BIR` is the endpoint for Birmingham City Council in the `local-authority-eng` register.
+
+| Endpoint | Description |
```
bahmady: Would it actually be more helpful to have a three-column table with "Method" "HTTP request" and "Description"?

source/documentation/getting_started/quick_start.md
arnau: I agree with `field-value` not being clear enough.  I think though, it is a job that needs a review across the whole documentation and probably out of scope for this PR.

bahmady:
Thanks @danmitchell- —I've added quite a few comments so between mine and yours there's now a fair bit to review in this PR.

source/documentation/getting_started/quick_start.md
arnau: if it helpls users frame the type of API, I don't see why not.

arnau:
After a conversation with @bahmady, I don't see how the table in (2) helps a user get a sense of what can be done with the API.  I think a list of operations (perhaps like the description column) seems to give enough information.  Then, each item can link to the section explaining the technical details: HTTP method, path, parameters, options to consider, etc.

What do you think?

nacnudus:
Maybe I'm weird but I prefer to read the second column of the table in (2), rather than the column with the URL templates.  I would also click on links to further explanation of "item", "entry" and "record" if they existed, because this is the point at which I realise that I don't know which one I want.

bahmady:
I'm inclined to agree with @arnau ... I'll open up a new commit now and ask for feedback.

source/documentation/getting_started/quick_start.md
bahmady: I did check with Conor but he wasn't entirely sure going forwards what we'll be calling it, so latest PR just changes it to "here".

source/documentation/getting_started/quick_start.md
bahmady: Yes, a bit careless of me re. "key", sorry. Latest commit for this specific PR addresses the backticks around "field-value".

source/documentation/getting_started/quick_start.md
```diff
@@ -1,10 +1,59 @@
 ## Quick start guide
 
-Registers in beta are reliable and are ready for use in products and services.
+Registers which are in beta are reliable and ready for use in products and services. Each register has an open, RESTful API which you can use to access the data in the register.
```
arnau: Is the "are in beta" helping clarify that it is the same as "ready to use"? Or it's something to remove?

source/documentation/getting_started/quick_start.md
```diff
@@ -1,10 +1,59 @@
...
 
-1. [Find the base URL for the register(s) you want to use](https://registers.cloudapps.digital/registers). For example, for the `local-authority-eng` register: `https://local-authority-eng.register.gov.uk/`.
-2. [Find the endpoints you want](#apireference). For example, `/record/BIR` in `https://local-authority-eng.register.gov.uk/record/BIR`.
-3. Add `.json`, `.yaml`, `.csv`, `.tsv`, or `.ttl` to the end of the request URL (before any query parameters) to choose the format of the response, or amend headers to `accept application/json` or equivalent.
-4. Parse register data and use it in your product or service.
+You can follow this guide to quickly get started with GOV.UK Registers APIs. 
+
+**1. Find the base URL for the register(s) you want to use.** 
+
+You can find a full list of register base URLs [here](https://registers.cloudapps.digital/registers).
```
arnau: I'm probably picky here but, is the "here" link a good practice in terms of accessibility? Back in the day links had to aim to be meaningful without contextual information so in this case the link would've been "list of register base URLs"

source/documentation/getting_started/quick_start.md
```diff
@@ -1,10 +1,59 @@
...
+
+The following are some examples of base URLs:
+
+| Register name | Base URL |
+|----------|----------|
+| `local-authority-eng`     | `https://local-authority-eng.register.gov.uk/`|
+| `country` | `https://country.register.gov.uk/` |
+| `allergen`  | `https://allergen.register.gov.uk/` |
```
arnau: 👌

source/documentation/getting_started/quick_start.md
danmitchell-: Aside from the content the linked page doesn't show "base URL's"

source/documentation/getting_started/quick_start.md
```diff
@@ -1,10 +1,59 @@
...
-4. Parse register data and use it in your product or service.
+You can follow this guide to quickly get started with GOV.UK Registers APIs. 
+
+**1. Find the base URL for the register(s) you want to use.** 
+
+You can find a full list of register base URLs [here](https://registers.cloudapps.digital/registers).
+
+The following are some examples of base URLs:
+
+| Register name | Base URL |
```
arnau: Technically this is the "Register ID" not the "Register name". Once the work done by @michaelabenyohai gets merged it will be more obvious.

E.g. 
Name: Local Authorities in England
ID: local-authority-eng

source/documentation/getting_started/quick_start.md
```diff
@@ -1,10 +1,59 @@
...
+* [get all records from a register](#get-records) 
+* [find a specific record within a register](#get-record-field-value) 
+* [get all entries for a single record](#get-record-field-value-entries) 
+* [find all records that share a `field-value` for a particular field](#get-records-field-name-field-value) 
+* [get all entries from a register](#get-entries)
+* [find a specific entry from a register](#get-entry-entry-number)
+* [find a specific item within a register](#get-item-item-hash)
+* [download the full contents of a register in a ZIP file](#get-download-register) 
+
+For example, `GET record/BIR` is the endpoint for Birmingham City Council in the `local-authority-eng` register, and 'BIR' is the `field-value`.
```
arnau: it should be `GET /record/BIR` to fully express that the path is absolute.

source/documentation/getting_started/quick_start.md
```diff
@@ -1,10 +1,59 @@
...
+
+**3. Choose the format of the response.**
+
+You can choose from the following response formats:
+
+* `.json` (JSON)
+* `.yaml` (YAML) 
+* `.csv` (CSV) 
+* `.tsv` (TSV)
+* `.ttl` (TTL)
```
arnau: Do we have elsewhere the full table with format, suffix and media type?

format|suffix|media type
------ | ---- | -----------
JSON|.json|application/json
YAML|.yaml|not registered, what are we accepting here?
CSV|.csv|text/csv
TSV|.tsv|text/tsv
Turtle|.ttl|text/ttl

arnau: The spec says `text/yaml` http://openregister.github.io/specification/#yaml-representation

arnau (changes requested):

source/documentation/getting_started/quick_start.md
nacnudus: Is this line still needed given the bullet points above?

source/documentation/getting_started/quick_start.md
```diff
@@ -1,10 +1,59 @@
...
+
+You can also specify a format by making a request with different headers. For example, this is also a valid request:
+
+```
+curl https://country.register.gov.uk/record/GB --header 'accept: application/json'
+```
+
+**4. Parse register data and use it in your product or service.**
 
 Most of the API functions return paginated results. Follow the specific guidance for each endpoint to get the page(s) of entries, items, and records you need. Use `page-size` to define the number of results you want and `page-index` to define the pages you want. The maximum `page-size` is 5000.
```
nacnudus: Maybe this should link to [#get-records](#get-records) which is the only endpoint with an example of `page-size` and `page-index`.

source/documentation/getting_started/quick_start.md
bahmady: It's a holdover from old language I didn't write, so I would look to you to tell me if we should just remove it.

source/documentation/getting_started/quick_start.md
bahmady: @arnau Yes, this was just a case of me trying to cram in too much last night. You're right.  @danmitchell-'s correct as well. I'm inclined to think it would actually be clearer to get rid of this altogether and for me to write a new sentence which explains it better from context, without giving a list of base URLs.

arnau: 👍 to @danmitchell- suggestion

source/documentation/getting_started/quick_start.md
bahmady: In this instance I can't think of a way to restructure this sentence since "beta" is such an important part of it:

> Registers which are in beta are reliable and ready for use in products and services.

So perhaps just delete this sentence?

arnau: I'm interested in other opinions, @michaelabenyohai @danmitchell-

source/documentation/getting_started/quick_start.md
bahmady: The idea of putting this in was to reinforce the point by example, since previous user research indicated that people struggled with the descriptions surrounding "endpoint".

source/documentation/getting_started/quick_start.md
bahmady: I've put together a new table which will be in the next commit 👍

source/documentation/getting_started/quick_start.md
bahmady: After discussing this with @arnau we've rejigged this section and it'll be in the new commit. It will need revisiting in a future PR as well.

source/documentation/getting_started/quick_start.md
bahmady: Even if it's true I think it's not really necessary to put it in the API docs; it's the sort of information that really belongs in more of a frontend context IMO. I've removed the sentence in the latest commit but depending on further feedback I can revisit/reintroduce it.

source/documentation/getting_started/quick_start.md
nacnudus: It's quite far out of context now though.  This particular example is for the third of the endpoints above, and it's the only mention of `field-value`.  I think users would be better of following the links above to find out how to use a given endpoint.

arnau (approved):

bahmady:
Would be great to get another review of the PR as it stands since it's quite an important change, but since it would be equally great to have this pushed in some state before Emma's UR, I will merge it before then.

source/documentation/getting_started/quick_start.md
bahmady: Yes, looking at it again after all the other changes, I agree with you. cc @arnau ?

arnau (approved):

dankmitchell (approved):
Nice

nacnudus (approved):

michaelabenyohai (approved):
👍

PR #35, opened 2018-04-05
Add API key flow information

subatoi:
Starter for ten on the API key flow

### DIFF
--- a/source/documentation/api_key/api_key.md
+++ b/source/documentation/api_key/api_key.md
@@ -1,8 +1,8 @@
-## API keys
+## Authentication 
 
 If you have created an API key then you can use it with all registers.
 
-Your API key should be of the form `76415efc-3d0f-4522-affd-bb6464e5ea88`. You should it in requests as part of the `Authorization` header. For example, this is a valid request:
+Your API key should be of the form `76415efc-3d0f-4522-affd-bb6464e5ea88`. You should place it in requests as part of the `Authorization` header. For example, this is a valid request:
 
 ```
 curl https://country.register.gov.uk/record/GB.json -H "Authorization: 76415efc-3d0f-4522-affd-bb6464e5ea88"
--- a/source/documentation/api_key/api_key.md
+++ b/source/documentation/api_key/api_key.md
@@ -1,0 +1,9 @@
+## API keys
+
+If you have created an API key then you can use it with all registers.
+
+Your API key should be of the form `76415efc-3d0f-4522-affd-bb6464e5ea88`. You should it in requests as part of the `Authorization` header. For example, this is a valid request:
+
+```
+curl https://country.register.gov.uk/record/GB.json -H "Authorization: 76415efc-3d0f-4522-affd-bb6464e5ea88"
+```
--- a/manifest.yml
+++ b/manifest.yml
@@ -1,6 +1,6 @@
 ---
 applications:
-- name: registers-docs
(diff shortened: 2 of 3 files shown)


source/documentation/api_key/api_key.md
```diff
@@ -0,0 +1,9 @@
+## API keys
```
arnau: I think that "API keys" is not the primary topic here. I think this should be "Authentication" (or "Authorisation", I think I prefer the first). As part of this section we can explain how to obtain an API key (or a link to the process, or both) and then how to use it (what you've written here).

Up for discussion but might be a good opportunity to highlight that although we don't currently restrict anything to anonymous requests, we reserve the right to do so in the future e.g. for rate limiting and that we encourage anyone to create one.

arnau (changes requested):

source/documentation/api_key/api_key.md
danmitchell-: https://docs.payments.service.gov.uk/#quick-start-guide

danmitchell-: I agree with @arnau. IMO this content should be the first step to the "quick start section"

source/documentation/api_key/api_key.md
```diff
@@ -0,0 +1,9 @@
+## Authentication 
+
+If you have created an API key then you can use it with all registers.
+
+Your API key should be of the form `76415efc-3d0f-4522-affd-bb6464e5ea88`. You should place it in requests as part of the `Authorization` header. For example, this is a valid request:
```
nacnudus: Just checking this isn't a real key!

source/documentation/api_key/api_key.md
```diff
@@ -0,0 +1,9 @@
+## Authentication 
+
+If you have created an API key then you can use it with all registers.
```
nacnudus: Will this eventually link to the process of creating a key?

source/documentation/api_key/api_key.md
```diff
@@ -0,0 +1,9 @@
...
+
+Your API key should be of the form `76415efc-3d0f-4522-affd-bb6464e5ea88`. You should place it in requests as part of the `Authorization` header. For example, this is a valid request:
+
+```
+curl https://country.register.gov.uk/record/GB.json -H "Authorization: 76415efc-3d0f-4522-affd-bb6464e5ea88"
```
michaelabenyohai: We use `--header` rather than `-H` in all other examples so just wondering whether we should be consistent. I don't mind which one we pick.

source/documentation/api_key/api_key.md
bahmady: No, it's from a GUID generator

gidsg: I would not include a realistic API key here as people might just use that instead of registering. I would rather say something like:
> Use the API key you received when registering

and then in the curl example do:

`curl https://country.register.gov.uk/record/GB.json -H "Authorization: YOUR-API-KEY-HERE"`

bahmady: @gidsg good shout but I've closed this PR; the new one is https://github.com/alphagov/registers-tech-docs/pull/43

gidsg: I would rather explicitly say:
> It should be included in all requests using the `Authorization` header

PR #43, opened 2018-04-09
New quick start guide format to add API key flow

bahmady:
This replaces https://github.com/alphagov/registers-tech-docs/pull/35#pullrequestreview-109971492


source/documentation/getting_started/quick_start.md
```diff
@@ -1,8 +1,26 @@
-## Quick start guide
+##Quick start guide
 
 Each register has an open, RESTful API which you can use to access the data in the register. 
 
-**1. Find the base URL for the register(s) you want to use.** 
+###Generate an API key 
+
+Visit the "Use the API" tab on the register you want to use and click **Create API key**.
```
gidsg: I would just link directly e.g.
```
(Create an API Key)[https://registers-trial.service.gov.uk/api_users/new] 
```
I would hope it it'd be self describing from there until...

> Your API key should be included in all requests using the `Authorization` header:

PR #44, opened 2018-04-09
Add Docker setup

arnau:
Defines a self contained Docker image and a complementary Docker Compose
configuration to simplify commands.

Caveat: Livereload doesn't work because it requires configuring the ws
host when the plugin gets activated. It's not possible with the current
setup.


arnau:
I changed this PR according to features available in tech docs gem 1.2+ so it should work with livereload as well.

PR #48, opened 2018-04-13
Remove api key flow

bahmady:
Reverts commits which concern API key information


gidsg:
I don't think this https://github.com/alphagov/registers-tech-docs/commit/c15f64ab5d185e768ac17ea67ce752a1d369191d has anything API key related so I don't think it needs reverting?

bahmady:
Yes, you're right ... I'll open a new PR then.

PR #51, opened 2018-04-16
Change desc of headers

subatoi:
Specify use of the `accept` header as opposed to more vague language:

"You can also specify a format by making a request with the `accept` header."


gidsg:
looks fine to me, but can you take a look @arnau

arnau (commented):
It would be perfect with the header `Accept` capitalised :)

bahmady:
Presumably we should be capitalising all such instances then?

arnau (approved):

PR #58, opened 2018-05-04
Fix seo title

danmitchell-:
GOV.UK Registers instead of GDS Registers


michaelabenyohai:
Any idea why it has "| Registers on the end of it"? I.e. it's "GOV.UK Registers | Registers". Should we just have "GOV.UK Registers" alone?

danmitchell-:
@michaelabenyohai that comes from the tech docs gem template so we'd have to make a PR against that to update it.

PR #61, opened 2018-05-11
Move 'Getting updates' section to own  file

subatoi:
Makes in line with rest of tech docs and easier to keep track of future changes to this content

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -158,7 +158,7 @@ PLATFORMS
   ruby
 
 DEPENDENCIES
-  govuk_tech_docs
+  govuk_tech_docs (~> 1.3)
   therubyracer
   tzinfo-data
   wdm (~> 0.1.0)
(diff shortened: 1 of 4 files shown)


Gemfile.lock
```diff
@@ -158,7 +158,7 @@ PLATFORMS
   ruby
 
 DEPENDENCIES
-  govuk_tech_docs
+  govuk_tech_docs (~> 1.3)
```
gidsg: not sure what's going on here but `Gemfile.lock` should not change unless `Gemfile` is changed and you re-run bundle install.

bahmady: I think this must be because I re-ran `bundle install` after the recent updates to the tech-docs-template

arnau: I also pinned this version in `Gemfile` (now in master). I have the feeling it was that change I made that was half way and didn't update the `Gemfile.lock` accordingly.

arnau (approved):

PR #62, opened 2018-05-11
Introduces item/entry/record content

subatoi:
Introduces this content and a worked example

PR #64, opened 2018-05-14
Change instances of 'field-value' to 'key' where appropriate

subatoi:
Assuming this is done properly, my feeling is we should use the term "key" rather than "primary key" (as long as this is valid)


bahmady:
Depending on the use of "primary key" vs "key" this would also affect https://github.com/alphagov/registers-tech-docs/pull/62

dankmitchell (commented):
I like this change

arnau (approved):

PR #67, opened 2018-05-18
Attempt to tidy Git up

bahmady:
I accidentally seem to have created a 'ghost' directory at some point but it doesn't seem to have been a full submodule. An empty `registers-tech-docs` file within the parent one.

Also, @danmitchell-  I've removed some references to some branches that I don't think exist any more, in the `.git/config` file; please let me know if I shouldn't do this.


danmitchell-:
`.git/config` is local your machine. Not in source control so we're all good 👍

PR #68, opened 2018-05-18
Add authorization header to all curl examples

michaelabenyohai:
We should encourage usage of the API key in all examples.

### DIFF
--- a/source/documentation/getting_started/quick_start.md
+++ b/source/documentation/getting_started/quick_start.md
@@ -10,7 +10,7 @@ Your API key can be used for all registers.
 
 You should include your key in all requests using the `Authorization` header:
 
-`curl https://country.register.gov.uk/record/GB.json --header "Authorization: YOUR-API-KEY-HERE"`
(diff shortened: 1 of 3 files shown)


source/documentation/getting_started/quick_start.md
```diff
@@ -10,7 +10,7 @@ Your API key can be used for all registers.
 
 You should include your key in all requests using the `Authorization` header:
 
-`curl https://country.register.gov.uk/record/GB.json --header "Authorization: YOUR-API-KEY-HERE"`
+`curl https://country.register.gov.uk/record/GB.json --header 'Authorization: YOUR-API-KEY-HERE'`
```
gidsg: this should be `/records/` now.

bahmady: @gidsg is correct, though I think (fortunately) this is the only example where this is incorrect

michaelabenyohai: ah good spot!

PR #74, opened 2018-05-25
Entry timestamp (suggestion to update existing other content)

bahmady:
The primary goal is to introduce entry timestamp content but I wasn't sure if it made sense to introduce that in isolation, which is why I've suggested this rework of existing content. If it would be more appropriate in isolation, I'll open a new PR.


bahmady:
Another question I had was whether we need the content about `index-entry-number` here

source/documentation/using_the_api/item_entry_record.md
```diff
@@ -0,0 +1,31 @@
...
+
+Using the `GET /records/{key}` endpoint returns the most recent entry to the register for a given `{key}`, with the item data already resolved. For example, `GET /records/CI` returns the [207th entry and its corresponding item]( https://country.register.gov.uk/records/CI.json). 
+
+Using the `GET /records/CI/entries` endpoint [returns all entries for the key 'CI'](https://country.register.gov.uk/records/CI/entries.json), meaning both the 90th and the 207th entries. 
+
+### The components of an entry 
+
+Each entry in a register is made up of the fields `entry-number`, `index-entry-number`, `entry-timestamp`, `item-hash` and `key`. The `entry-number` is unique and defines an entry’s position within the ordered list of a register. The `index-entry-number` is unique and defines an entry’s position within the ordered list of an index. For an entry in a register the `entry-number` and `index-entry-number` are always identical.
```
bahmady: Do we need to document `index-entry-number` here?

PR #75, opened 2018-06-01
Datatypes

subatoi:
First draft of introduction of content about datatypes (how to resolve foreign keys, CURIEs). I think this needs some context surrounding what the user need is.


source/documentation/getting_started/quick_start.md
```diff
@@ -16,6 +16,8 @@ You should include your key in all requests using the `Authorization` header:
 
 Find the base URL for each register on its API inspector. For example, the API inspector for the `local-authority-eng` register is located at `https://local-authority-eng.register.gov.uk/`. 
 
+The base URL of a register is always of the form `https://{register}.{domain}`. For example, the base URL for the `local-authority-eng` register is `https://local-authority-eng.register.gov.uk`, so the domain for this register is `register.gov.uk`.
```
bahmady: Rather than use the `datatypes.md` file to have this content as part of a tutorial, I've suggested expanding the content about base URLs in the 'Quick start guide' section. This may not be the best way, so I am open to alternative suggestions: this avoids the problem of having two 'For example' structures in quick succession in `datatypes.md` (i.e. example about what a domain is, then example about how to find out what the datatype is).

source/documentation/using_the_api/datatypes.md
```diff
@@ -0,0 +1,26 @@
+## Datatypes
```
bahmady: I think this would be better if it were centred around a 'How to', as the CURIEs and foreign keys sections are, but the reason I've written it this way is to accommodate GOV.UK style limitations about numbered lists. I'm not sure if the focus should be on resolving foreign keys and CURIEs, though, in which case this should be more of a 'Prerequisites' section.

bahmady: I called this 'Datatypes' because it was much more straightforward to make the content GOV.UK-appropriate and to fit it in with the rest of the documentation rather than having it be focused on CURIEs and foreign keys and have ascertaining the domain/datatype be a prerequisite. If this approach doesn't make sense just let me know.

source/documentation/using_the_api/datatypes.md
```diff
@@ -0,0 +1,26 @@
...
+
+The simplest way to find a field’s datatype is to use the record in the appropriate field register. This is of the form `https://field.{domain}/records/{field}`, where `{domain}` is the [domain of your register from its base URL](#find-the-base-url-for-register-s
```
arnau: Currently, the only field register is https://field.register.gov.uk/ so wouldn't it be better to remove the variability of `domain`?

source/documentation/using_the_api/datatypes.md
```diff
@@ -0,0 +1,26 @@
...
+), and `{field}` is the field you’re interested in. 
+
+For example, the [datatype of the `local-authority-type` field in the `local-authority-eng` register is `string`](https://field.register.gov.uk/records/local-authority-type.json). 
+
+Go to `https://datatype.{domain}/records/{datatype}` to find the definition of any given datatype of interest (`{datatype}`).
```
arnau: Same as with the field register comment.

source/documentation/using_the_api/datatypes.md
```diff
@@ -0,0 +1,26 @@
+## Datatypes
+
+The simplest way to find a field’s datatype is to use the record in the appropriate field register. This is of the form `https://field.{domain}/records/{field}`, where `{domain}` is the [domain of your register from its base URL](#find-the-base-url-for-register-s
+), and `{field}` is the field you’re interested in. 
+
+For example, the [datatype of the `local-authority-type` field in the `local-authority-eng` register is `string`](https://field.register.gov.uk/records/local-authority-type.json). 
+
+Go to `https://datatype.{domain}/records/{datatype}` to find the definition of any given datatype of interest (`{datatype}`). 
+
+### How to resolve foreign keys
```
arnau: I don't think this section belongs to "Datatypes"

source/documentation/using_the_api/datatypes.md
```diff
@@ -0,0 +1,26 @@
...
+
+A field is a foreign key to another register if the `register` field in the record of the field register (`https://field.{domain}/records/{field}`) is populated. 
+
+For example, the `local-authority-type` field is [a foreign key to the `local-authority-type` register](https://field.register.gov.uk/records/local-authority-type.json). `NMD` populates [the `SHE` record of this register](https://local-authority-eng.register.gov.uk/records/SHE.json), so this resolves to [the `NMD` record in the `local-authority-type` register](https://local-authority-type.register.gov.uk/records/NMD.json).  
+
+### How to resolve CURIEs
```
arnau: I don't think this section belongs to "Datatypes"

source/documentation/using_the_api/datatypes.md
```diff
@@ -0,0 +1,26 @@
...
+
+Any value in a field with a datatype of `curie` is a CURIE. In registers, all CURIEs are of the form `https://{register}.{domain}/records/{key}`.
```
arnau: Let's chat about this one

arnau (changes requested):

source/documentation/using_the_api/datatypes.md
bahmady: I think @michaelabenyohai was accounting for `alpha.openregister.org` as well. If we only have one domain it would make the documentation considerably easier to write.

source/documentation/using_the_api/datatypes.md
bahmady: Yes, I thought it may not (same for the CURIEs section); that's why I added a pre-emptive comment under the 'Datatypes' h2.

PR #76, opened 2018-06-04
Entry timestamp content

subatoi:
The primary goal originally was simply to introduce entry timestamp content but I wasn't sure if it made sense to introduce that in isolation, which is why I've suggested this rework of existing content. If it would be more appropriate in isolation, I'll open a new PR.


source/documentation/using_the_api/register_components.md
```diff
@@ -1,18 +1,30 @@
...
-Each entry in a register has a relationship with one item, which contains data. An item is connected to an entry with an item hash.   
+Entries are made up of the fields `entry-number`, `index-entry-number`, `entry-timestamp`, `item-hash` and `key`. 
+
+The `entry-number` is unique and defines an entry’s position within the ordered list of a register. The `index-entry-number` is unique and defines an entry’s position within the ordered list of an index. For any given entry in a register, the `entry-number` and `index-entry-number` are always identical. 
+
+The `entry-timestamp` is the time when an entry was introduced to a given register, but it is no guarantee of the order of entries in a register. The `entry-number` is.
+
+### Items
+
+Each entry in a register has a relationship with one item. Items contain data for a given register. Entries and items are connected by a given `item-hash`.
```
arnau: I wonder if we could flip the sentences so the definition of what is an item comes first. Also, it would be interesting to hint that an item is the recipient for a structured set of data where the fields are pre-defined and consistent within a register.

source/documentation/using_the_api/register_components.md
```diff
@@ -1,18 +1,30 @@
...
+### Entries
+
 Entries contain metadata about what and when data has changed. 
 
-Each entry in a register has a relationship with one item, which contains data. An item is connected to an entry with an item hash.   
+Entries are made up of the fields `entry-number`, `index-entry-number`, `entry-timestamp`, `item-hash` and `key`. 
+
+The `entry-number` is unique and defines an entry’s position within the ordered list of a register. The `index-entry-number` is unique and defines an entry’s position within the ordered list of an index. For any given entry in a register, the `entry-number` and `index-entry-number` are always identical. 
+
+The `entry-timestamp` is the time when an entry was introduced to a given register, but it is no guarantee of the order of entries in a register. The `entry-number` is.
```
arnau: I think `key` should be described as well. And probably explain it has the same value as the identifier (primary key) in the item.

source/documentation/using_the_api/register_components.md
```diff
@@ -1,18 +1,30 @@
...
 * records
 
+### Entries
+
 Entries contain metadata about what and when data has changed. 
 
-Each entry in a register has a relationship with one item, which contains data. An item is connected to an entry with an item hash.   
+Entries are made up of the fields `entry-number`, `index-entry-number`, `entry-timestamp`, `item-hash` and `key`. 
+
+The `entry-number` is unique and defines an entry’s position within the ordered list of a register. The `index-entry-number` is unique and defines an entry’s position within the ordered list of an index. For any given entry in a register, the `entry-number` and `index-entry-number` are always identical.
```
arnau: @michaelabenyohai given that indexes are not a stable feature I'm thinking that the description of `index-entry-number` should be separated from the `entry-number` with the note/warning that it is not ready to be used by regular users.

source/documentation/using_the_api/register_components.md
arnau: > Items contain data for a given register.

Would it be more useful if we say that an item contains data for a _thing_ in a register? Perhaps I'm overthinking it.

source/documentation/using_the_api/register_components.md
```diff
@@ -1,18 +1,30 @@
...
+The `entry-timestamp` is the time when an entry was introduced to a given register, but it is no guarantee of the order of entries in a register. The `entry-number` is.
+
+### Items
+
+Each entry in a register has a relationship with one item. Items contain data for a given register. Entries and items are connected by a given `item-hash`.  
+
+### Records
 
-Records represent things that change over time. Records are identifiable by keys and correspond to the latest entry for a particular key, as indicated by the entry number.
+Records represent things that change over time. Records are identifiable by keys and correspond to the latest entry for a particular key, as indicated by a given `entry-number`.
```
arnau: > Records represent things that change over time

I'm not sure I agree with this statement. A record represents a sigle thing given a register in a particular state.

arnau (changes requested):

source/documentation/using_the_api/register_components.md
bahmady: I agree with all of these points, to varying extents. Item first is no problem; I just put entry first because it was the biggest. Describing an item as a structured set of data is also no problem. The only thing I'm not going to propose in the next commit will be to say that items contain data for things in registers, because I've reworked the definition around the first point:

> Items contain structured data sets. They have predefined fields which are consistent within a register.

I think I like this definition because it strays away from an abstract definition of what items are and sticks to a rigorous technical definition. 

That being said if it gets tested and people struggle with it then of course ...

source/documentation/using_the_api/register_components.md
bahmady: Makes sense to me. We should include a definition though, as it's in all the [example] outputs.

source/documentation/using_the_api/register_components.md
bahmady: It's too vague, I agree. Is it fair to relate records explicitly to keys, in a definition?

source/documentation/using_the_api/register_components.md
bahmady: Yes, this is actually an oversight (not describing `key`). Would this be a good opportunity to introduce foreign keys, juxtaposed against primary keys... something like this:

The `key` is a numeric or alphanumeric string. The primary key always has the same name as the name of the register. For example, every entry in the `local-authority-eng` register has a field called `local-authority-eng`. [Foreign keys, which are not always present, link to other registers](link-to-content-about-foreign-keys).

source/documentation/using_the_api/register_components.md
arnau: I like the reworked definition, let's see how it tests out.

Just to clarify, when I said "flip the sentences so the definition of what is an item comes first" I was talking about the paragraph within the "Items" section, not at all referring to flipping entry and item sections.

source/documentation/using_the_api/register_components.md
arnau: > The key is a numeric or alphanumeric string

"The key is an alphanumeric string" is better given that even if it only contains numbers it fits the definition. The point is that you must not treat it as a number ever.

arnau: I like how you introduce the foreign key concept. If possible, it would be good to equal "primary key" to "unique identifier". Mainly because ID is a concept more popular in other HTTP APIs and at some point I would like to use "id" as the field name.

source/documentation/using_the_api/register_components.md
arnau: Sure, as long as it is clear that no one should rely on it. Also, the concept of "index" is not used at the user level right now so mentioning here is not that useful, I think.

source/documentation/using_the_api/register_components.md
arnau: I'm ok with the rewrite.

source/documentation/using_the_api/register_components.md
bahmady: Hmm, perhaps a separate PR then?

source/documentation/using_the_api/register_components.md
bahmady: I was also wondering about "alphanumeric": the definitions I've found seem to suggest that "alphanumeric" means "a mix of letters and numbers" whereas "numeric" means "just numbers". I'm inclined to agree with you, but it's an interesting question.

On the second point, I'll rework it such that it implies an equivalence with primary key and unique identifier.

source/documentation/using_the_api/register_components.md
bahmady: @arnau Ah, I see. Yes, that makes sense: the definition of an item before was cyclical, and relied on entries to make sense. That wasn't ideal; I think it's better now. 

@nacnudus How about "items contain data for a given entry in a register"?

nacnudus: @bahmady That's fine with me :)

source/documentation/using_the_api/register_components.md
```diff
@@ -1,18 +1,34 @@
...
+### Entries
+
 Entries contain metadata about what and when data has changed. 
 
-Each entry in a register has a relationship with one item, which contains data. An item is connected to an entry with an item hash.   
+Entries contain the fields `entry-number`,  `entry-timestamp`, `item-hash` and `key`. 
+
+The `entry-number` is unique and defines an entry’s position within the ordered list of a register. The `entry-timestamp` is the time when an entry was introduced to a given register, but it is no guarantee of the order of entries in a register. The `entry-number` is.
+
+Entries are connected to items by a given `item-hash`.
```
arnau: Being picky: Entries reference items in the `item-hash` field using the hash derived from the item content.

What I think it's important is to not confuse "`item-hash`" and item hash. The former is a field name, the latter is the actual alphanumeric value.

source/documentation/using_the_api/register_components.md
```diff
@@ -1,18 +1,34 @@
...
 Entries contain metadata about what and when data has changed. 
 
-Each entry in a register has a relationship with one item, which contains data. An item is connected to an entry with an item hash.   
+Entries contain the fields `entry-number`,  `entry-timestamp`, `item-hash` and `key`. 
+
+The `entry-number` is unique and defines an entry’s position within the ordered list of a register. The `entry-timestamp` is the time when an entry was introduced to a given register, but it is no guarantee of the order of entries in a register. The `entry-number` is.
+
+Entries are connected to items by a given `item-hash`. 
+
+Each `key` is an alphanumeric string unique to the data in a register. The primary key (i.e. the unique identifier) always has the same name as the name of the register. For example, every entry in the `local-authority-eng` register has a field called `local-authority-eng`. Foreign keys, which are not always present, link to other registers.
```
arnau: I realised we have keys that contain other carachters than just alphanumeric. I need to see how many registers are afected by this. /cc @gidsg @nacnudus

source/documentation/using_the_api/register_components.md
```diff
@@ -1,18 +1,34 @@
...
-Each entry in a register has a relationship with one item, which contains data. An item is connected to an entry with an item hash.   
+Entries contain the fields `entry-number`,  `entry-timestamp`, `item-hash` and `key`. 
+
+The `entry-number` is unique and defines an entry’s position within the ordered list of a register. The `entry-timestamp` is the time when an entry was introduced to a given register, but it is no guarantee of the order of entries in a register. The `entry-number` is.
+
+Entries are connected to items by a given `item-hash`. 
+
+Each `key` is an alphanumeric string unique to the data in a register. The primary key (i.e. the unique identifier) always has the same name as the name of the register. For example, every entry in the `local-authority-eng` register has a field called `local-authority-eng`. Foreign keys, which are not always present, link to other registers. 
+
+> Entries also contain the field `index-entry-number`, which is a unique number that defines an entry's position within the ordered list of an index. Indexes are not a reliable feature yet, which are not ready for use by most users.
```
arnau: What about putting the warning in another paragraph, on the lines of:

> **Note**: `index-entry-number` depends on indexes, an experimental feature. Use it at your own risk.

source/documentation/using_the_api/register_components.md
```diff
@@ -1,18 +1,34 @@
...
+
+Entries are connected to items by a given `item-hash`. 
+
+Each `key` is an alphanumeric string unique to the data in a register. The primary key (i.e. the unique identifier) always has the same name as the name of the register. For example, every entry in the `local-authority-eng` register has a field called `local-authority-eng`. Foreign keys, which are not always present, link to other registers. 
+
+> Entries also contain the field `index-entry-number`, which is a unique number that defines an entry's position within the ordered list of an index. Indexes are not a reliable feature yet, which are not ready for use by most users.  
+
+### Items
+
+Items contain structured data for a given primary key in a register. Items have predefined fields which are consistent within a register.
```
arnau: Did you decide to not mention identifiers?

arnau (changes requested):

source/documentation/using_the_api/register_components.md
bahmady: @arnau 

> Did you decide to not mention identifiers?

I tried to mention it earlier in the text, in the context of entries:

> Each `key` is an alphanumeric string unique to the data in a register. The primary key (i.e. the unique identifier) always has the same name as the name of the register. For example, every entry in the `local-authority-eng` register has a field called `local-authority-eng`. Foreign keys, which are not always present, link to other registers. 

> ... alphanumeric string _unique_ to the data ...
> ... (i.e. the unique identifier) ...

We could say instead(?):

> ... The primary key is a unique identifier for the register, and always has the same name as the name of the register ...

I think we need to be specific that this would be the primary key here? This old definition was floating around and I don't think this is accurate:

> Records define themselves by a unique identifier associated with a single record in a register. A unique identifier can be a candidate for a primary key. The primary key always has the same name as the name of the register. For example, every entry in the country register has a field called country. This helps to link that entry specifically to the country register.

source/documentation/using_the_api/register_components.md
arnau: > Each entry is connected to an item by a given item hash. Entries reference items in the item-hash field using the hash derived from the corresponding item's content.

The above looks much better, thanks.

> [...] so correct me if I'm wrong but it's not just the item hash/item-hash that has that problem in that whole part of the text but rather it's all of them (e.g. entry number, entry timestamp).

Well, in the other cases is an abuse of language but seems clear that you talk about the value of the field, not the field name itself.

source/documentation/using_the_api/register_components.md
bahmady: Same here. Some of my understanding about primary keys was drawn from that, but talking to you the other day I realise that paragraph is quite problematic.

source/documentation/using_the_api/register_components.md
bahmady: > Well, in the other cases is an abuse of language but seems clear that you talk about the value of the field, not the field name itself.

OK, so I'll correct the other cases at the same time. I think in doing that we solve both problems at the same time, but the main problem is with this sentence:

> Entries are connected to items by a given item hash. 

Which I think we can fix with: 

> Entries are connected to items by a given item hash.

source/documentation/using_the_api/register_components.md
arnau: I would like to keep the alpha/beta labels to a minimum, we've been burnt before by multiple labels for different things.

I don't know what is the GOV.UK style for warnings but if it expresses clearly that the feature is not reliable because it is experimental, I don't mind.

bahmady: No problem. How about keeping it relatively simple for now:

> `index-entry-number` depends on indexes, which are an experimental and unreliable feature. 

If I have to change this after a writer review, I will do.

source/documentation/using_the_api/register_components.md
```diff
@@ -12,21 +12,23 @@ Entries contain metadata about what and when data has changed.
 
 Entries contain the fields `entry-number`,  `entry-timestamp`, `item-hash` and `key`. 
 
-The `entry-number` is unique and defines an entry’s position within the ordered list of a register. The `entry-timestamp` is the time when an entry was introduced to a given register, but it is no guarantee of the order of entries in a register. The `entry-number` is.
+The entry number is unique and defines an entry’s position within the ordered list of a register. The entry timestamp is the time when an entry was introduced to a given register, but it is no guarantee of the order of entries in a register. The entry number is.
```
arnau: Re-reading this paragraph, perhaps it's worth clarifying that the "ordered list" is actually the "ordered list of changes"? So no one confuses the list of entries with the list of records.

source/documentation/using_the_api/register_components.md
```diff
@@ -12,21 +12,23 @@ Entries contain metadata about what and when data has changed.
...
 
-Entries are connected to items by a given `item-hash`. 
+Each entry is connected to an item by a given item hash. Entries reference items in the `item-hash` field using the hash derived from the corresponding item's content.
 
-Each `key` is an alphanumeric string unique to the data in a register. The primary key (i.e. the unique identifier) always has the same name as the name of the register. For example, every entry in the `local-authority-eng` register has a field called `local-authority-eng`. Foreign keys, which are not always present, link to other registers. 
+Each key is an alphanumeric string unique to the data in a register. The primary key (i.e. the unique identifier) always has the same name as the name of the register. For example, every entry in the `local-authority-eng` register has a field called `local-authority-eng`. Foreign keys, which are not always present, link to other registers.
```
arnau: After reading the linking section in #77 I think there is no need to mention foreign keys here.

source/documentation/using_the_api/register_components.md
```diff
@@ -12,21 +12,23 @@ Entries contain metadata about what and when data has changed.
...
 
-Each `key` is an alphanumeric string unique to the data in a register. The primary key (i.e. the unique identifier) always has the same name as the name of the register. For example, every entry in the `local-authority-eng` register has a field called `local-authority-eng`. Foreign keys, which are not always present, link to other registers. 
+Each key is an alphanumeric string unique to the data in a register. The primary key (i.e. the unique identifier) always has the same name as the name of the register. For example, every entry in the `local-authority-eng` register has a field called `local-authority-eng`. Foreign keys, which are not always present, link to other registers. 
 
-> Entries also contain the field `index-entry-number`, which is a unique number that defines an entry's position within the ordered list of an index. Indexes are not a reliable feature yet, which are not ready for use by most users.  
+Entries also contain the field `index-entry-number`, which is a unique number that defines an entry's position within the ordered list of an index. 
+
+> index-entry-number depends on indexes, which are an experimental and unreliable feature.
```
arnau: This shouldn't be a blockquote, either a simple paragraph or one that shows it's a note or a warning.

arnau (changes requested):

source/documentation/using_the_api/register_components.md
```diff
@@ -12,15 +12,15 @@ Entries contain metadata about what and when data has changed.
...
+The entry number is unique and defines an entry’s position within the ordered list of changes. The entry timestamp is the time when an entry was introduced to a given register, but it is no guarantee of the order of entries in a register. The entry number is.
 
 Each entry is connected to an item by a given item hash. Entries reference items in the `item-hash` field using the hash derived from the corresponding item's content.
 
-Each key is an alphanumeric string unique to the data in a register. The primary key (i.e. the unique identifier) always has the same name as the name of the register. For example, every entry in the `local-authority-eng` register has a field called `local-authority-eng`. Foreign keys, which are not always present, link to other registers. 
+Each key is an alphanumeric string unique to the data in a register. The primary key (i.e. the unique identifier) always has the same name as the name of the register. For example, every entry in the `local-authority-eng` register has a field called `local-authority-eng`.
```
arnau: The `key` being or not alphanumeric is a longer conversation so this paragraph should change to something like:

> The key is a unique UTF-8 string identifying a _thing_ within a register. The name of the field holding this data in the item has the same name as the register. For example, every item in the `local-authority-eng` register has a field called `local-authority-eng`.

Besides removing the alphanumeric bit, I changed the example because the field with the same name as the register is in the item, not the entry.

arnau (changes requested):

source/documentation/using_the_api/register_components.md
```diff
@@ -1,18 +1,36 @@
...
 * entries
 * items
 * records
 
+### Entries
+
 Entries contain metadata about what and when data has changed. 
 
-Each entry in a register has a relationship with one item, which contains data. An item is connected to an entry with an item hash.   
+Entries contain the fields `entry-number`,  `entry-timestamp`, `item-hash` and `key`.
```
michaelabenyohai: Since the word "field" has a very specific meaning in the world of registers, I'm wondering whether we should we use the word "property" here? For example, these have nothing to do with the field register as fields in an item do. Thoughts @arnau?

source/documentation/using_the_api/register_components.md
```diff
@@ -1,18 +1,36 @@
...
 Entries contain metadata about what and when data has changed. 
 
-Each entry in a register has a relationship with one item, which contains data. An item is connected to an entry with an item hash.   
+Entries contain the fields `entry-number`,  `entry-timestamp`, `item-hash` and `key`. 
+
+The entry number is unique and defines an entry’s position within the ordered list of changes. The entry timestamp is the time when an entry was introduced to a given register, but it is no guarantee of the order of entries in a register. The entry number is.
+
+Each entry is connected to an item by a given item hash. Entries reference items in the `item-hash` field using the hash derived from the corresponding item's content.
+
+Each key is a unique UTF-8 string which identifies something in a register. The `key` field holding this data always has the same name as the register. For example, every entry in the `local-authority-eng` register has a field called `local-authority-eng`.
```
michaelabenyohai: This is not quite accurate. It is the **item** rather than the **entry** that has a field called `local-authority-eng`. There is a `key` property in the **entry**. The data contained in this property happens to be the same as the data contained in the field of the corresponding **item** that has the same name as the register.

Might be worth splitting it out into the corresponding entry and item sections?

Entry: 
> Each key is a unique UTF-8 string which identifies something in a register. 

Item:
> The item always contains a field which has the same name as the register. For example, every item in the `local-authority-eng` register has a field called `local-authority-eng`. The data contained in this field will be the same as the data in the `key` property of the entry that introduced the item to the register.

This second statement is not true for indexes, but I assume we're ignoring their existence for now.

michaelabenyohai (changes requested):
Looks good other than the couple of things I mentioned.

source/documentation/using_the_api/register_components.md
michaelabenyohai: Note that if we change this there's a few other places to change "field" to "property", unless we are actually talking about a field in the item.

source/documentation/using_the_api/register_components.md
bahmady: Sounds good to me. I'll move the content around as you suggested in the next commit.

bahmady:
@arnau is away until Wednesday I think so happy to wait until he's back to give him chance to reply to @michaelabenyohai's comments.

source/documentation/using_the_api/register_components.md
arnau: Sounds good to me, as you say, as long as we keep consistency across the docs.

bahmady: Would it make more sense to open a separate PR to cover all such instances/changes, in that case?

arnau: I think it's better to do it in a separate PR so you can focus on that specifically.

bahmady: Cool, are you happy to review the PR as it stands then?

arnau (approved):

michaelabenyohai (approved):

PR #77, opened 2018-06-07
Linked registers

subatoi:
Second version of content on CURIEs and foreign keys


source/documentation/using_the_api/linked_registers.md
```diff
@@ -0,0 +1,18 @@
+## Linked registers
+
+Registers can be linked by CURIEs and foreign keys.
```
arnau: Perhaps it is worth mentioning that CURIEs are the newest and preferred mechanism so new registers are expected to only use CURIEs.

source/documentation/using_the_api/linked_registers.md
```diff
@@ -0,0 +1,18 @@
+## Linked registers
+
+Registers can be linked by CURIEs and foreign keys. 
+
+### CURIEs 
+
+Any value in a field with a datatype of `curie` must be a CURIE. In registers, all CURIEs are of the form `{prefix}:{key}`, where `{prefix}` is mapped to a base URL `https://{register}.register.gov.uk/records/` for a given `{register}`. 
+
+For example, [the `organisation` field in the `statistical-geography` register](https://field.register.gov.uk/records/organisation.json) has the datatype `curie`. In the [‘E09’ record of this register](https://statistical-geography.register.gov.uk/records/E09.json), the value of the `organisation` field is `government-organisation:D4`, so this resolves to the [‘D4’ record in the `government-organisation` register](https://government-organisation.register.gov.uk/records/D4). In this example, the `{key}` is ‘D4’ and the `{prefix}` is `government-organisation`, which maps to the base URL `https://government-organisation.register.gov.uk/records/D4`.
```
arnau: What about something like:

> For example, [the `organisation` field in the `statistical-geography` register](https://field.register.gov.uk/records/organisation.json) has the datatype `curie`. In the [‘E09’ record of this register](https://statistical-geography.register.gov.uk/records/E09.json), the value of the `organisation` field is `government-organisation:D4`. Knowing that the prefix `government-organisation` maps the URL `https://government-organisation.register.gov.uk/records/` the expanded URL is `https://government-organisation.register.gov.uk/records/D4`.

source/documentation/using_the_api/linked_registers.md
```diff
@@ -0,0 +1,18 @@
...
+
+For example, [the `organisation` field in the `statistical-geography` register](https://field.register.gov.uk/records/organisation.json) has the datatype `curie`. In the [‘E09’ record of this register](https://statistical-geography.register.gov.uk/records/E09.json), the value of the `organisation` field is `government-organisation:D4`, so this resolves to the [‘D4’ record in the `government-organisation` register](https://government-organisation.register.gov.uk/records/D4). In this example, the `{key}` is ‘D4’ and the `{prefix}` is `government-organisation`, which maps to the base URL `https://government-organisation.register.gov.uk/records/D4`. 
+
+You can see the datatype of any given `{field}` at `https://field.register.gov.uk/records/{field}`. 
+
+### Foreign keys
+
+Some fields are foreign keys which link to other registers. You can check this for a given `{field}` using the field register (`https://field.register.gov.uk/records/{field}`).
```
arnau: Perhaps it's useful to explain that only fields with the `register` field informed are candidates to be foreign keys?

source/documentation/using_the_api/linked_registers.md
```diff
@@ -0,0 +1,18 @@
...
+
+You can see the datatype of any given `{field}` at `https://field.register.gov.uk/records/{field}`. 
+
+### Foreign keys
+
+Some fields are foreign keys which link to other registers. You can check this for a given `{field}` using the field register (`https://field.register.gov.uk/records/{field}`). 
+
+For example, the [`local-authority-type` field is a foreign key](https://field.register.gov.uk/records/local-authority-type.json). In the `local-authority-eng` register, the `local-authority-type` field for each record links to a corresponding record in the `local-authority-type` register. In [the ‘SHE’ record of the `local-authority-eng` register](https://local-authority-eng.register.gov.uk/records/SHE.json), ‘NMD’ populates the `local-authority-type` field, so this resolves to [the ‘NMD’ record in the `local-authority-type` register](https://local-authority-type.register.gov.uk/records/NMD.json).
```
arnau: Similarly to the CURIE example, I would avoid making links and instead I would show URLs as code.

arnau (changes requested):

source/documentation/using_the_api/linked_registers.md
nacnudus: `'NMD' populates the `local-authority-type` field`.  The active voice seems odd here, but this is nitpicking.

arnau (approved):

michaelabenyohai (approved):
Minor thing - all links in this example are to the HTML views whereas all links in the entries/items/records example link to the JSON views. Should we be consistent?

bahmady:
@michaelabenyohai probably less minor than you might think :) 

I actually removed the JSON extensions myself because my understanding is we're referring specifically to URLs in a pure sense in this particular doc. If that understanding's wrong then I agree it would be better to be consistent.

bahmady:
Apologies @michaelabenyohai I've been too close to too many rewrites and didn't realise there were still links that should have been JSON. I've made one such link a JSON view, consistent with the rest of the doc as you point out, and made another one an explicit URL as I think it helps readability.

bahmady:
Apologies, I didn't make this clear in the last commit but I rewrote some content slightly:

Under the 'Linked registers' header, I added:

> Foreign keys are fields, and not defined by a single datatype.

and under the 'Foreign keys' header, I changed some language:

> Some fields in registers are foreign keys which link to other registers. A given field (‘{field}’) is a foreign key to another register if the `register` field is populated at `https://field.register.gov.uk/records/{field}`.

michaelabenyohai:
@bahmady I guess I'm not sure what the correct thing to do is. 

In the entries/items example all full URLs are resolvable links to the JSON. We use code snippets for things like `GET /entries`, which I think is fine as this is not a full URL.

Where I find this inconsistent is that we are using code snippets for full URLs like `https://field.register.gov.uk/records/organisation`. The difference is, in the first case (entries/items) the code snippet version doesn't resolve to anything so that's fine. Whereas in this case, if a user were to copy and paste that into HTML and not JSON. I guess it depends whether we expect people to copy and paste this into the browser?

I'm happy for this to be merged and go to content review regardless. It's just something I find a little inconsistent.

bahmady:
@michaelabenyohai I see your point. What I'm wary of is that making the explicit URLs in code snippets resolvable links to the JSON might make it seem a bit confusing, given that the context for all the URLs in code snippets is "what [explicit] URL you would need", as in for example the context of {prefix}:{key} re. CURIEs. I don't think we necessarily mandate JSON or else we could just change them all. 

One thing we could do is make every URL that's an explicit URL _also_ a link, and have the link be the JSON view. That makes it much less likely that people will copy and paste, if they're just looking at the API docs.

bahmady:
Yeah that would be a fairly good indication of how well the content works, in itself.

bahmady:
Just an update (duplicated from Trello just in case): this is currently out for writer review as of 11/06.

arnau (approved):

PR #79, opened 2018-06-12
Deploy from Travis CI

gidsg:
* Add `.travis.yml` and change deploy script to use zero downtime plugin
* Switch to setting config in `Staticfile` as `nginx.conf` is deprecated in the buildpack

CC @J-Lambo


.travis.yml
```diff
@@ -0,0 +1,25 @@
+env:
+  global:
+  - CF_API="https://api.cloud.service.gov.uk"
+  - CF_ORG="openregister"
+  - CF_SPACE="docs"
+  - CF_USERNAME="register-design-authority@digital.cabinet-office.gov.uk"
```
dankmitchell: Really need to update this email but not important for this PR

PR #84, opened 2018-06-21
Reorganisation of API reference

subatoi:
Fairly significant reorganisation of the API reference information. Removes use of `curl` to make not only platform-agnostic but also tool-agnostic, and aligns better with specification. 

Should also be more developer-friendly. Feedback welcomed, as ever.

Having some problems with consistency in nomenclature and curly braces, for example with: 

> Get all records that share a `field-value` for a particular field-name

Also just to note that if this PR is approved in principle then we (I) will have to edit the quick start guide's links to match some new tags; the only reason I haven't done that yet is because I mainly want to use this to gather early feedback on this change for the time being. 

Please also see the notes below.

### DIFF
--- a/source/documentation/using_the_api/api_reference.md
+++ b/source/documentation/using_the_api/api_reference.md
@@ -1,8 +1,6 @@
 ## <a name="apireference">API reference</a>
 
-### <a name="get-register">View information about a register</a>
-
-Endpoint: `GET /register/`
+### <a name="getregister">`GET /register`</a>	
 
 ```http
 GET /register/ HTTP/1.1
@@ -38,9 +36,7 @@ Content-Type: application/json
 }
 ```
 
-### <a name="get-records">Get all records from a register</a>
-
-Endpoint: `GET /records/`
+### <a name="get-records">`GET /records/`</a>
 
 Parameters: 
 * `page-index` (Optional): Collection page number. Defaults to 1.
@@ -106,9 +102,7 @@ Link: <?page-index=2&page-size=3>; rel="next"
 }
 ```
 
-### <a name="get-records-key">Get a specific record within a register based on a particular key</a>
-
-Endpoint: `GET /records/{key}/`
+### <a name="get-records-key">`GET /records/{key}/`</a>
 
 Parameters: 
 * `{key}` (Required): A unique UTF-8 string which identifies something in a register.
@@ -142,9 +136,7 @@ Content-Type: application/json
 }
 ```
 
-### <a name="get-records-key-entries">Get all entries for a single record based on a particular key</a>
-
-Endpoint: `GET /records/{key}/entries/`
+### <a name="get-records-key-entries">`GET /records/{key}/entries/`</a>
 
 Parameters: 
 * `{key}` (Required): A unique UTF-8 string which identifies something in a register.
@@ -182,9 +174,7 @@ Content-Type: application/json
 ]
 ```
 
-### <a name="get-records-field-name-field-value">Get all records that share a `field-value` for a particular `field-name`</a>
-
-Endpoint: `GET /records/{field-name}/{field-value}/`
+### <a name="get-records-field-name-field-value">`GET /records/{field-name}/{field-value}/`</a>
 
 Parameters: 
 * `{field-name}` (Required): Field name. 
@@ -249,9 +239,7 @@ Link: <?page-index=2&page-size=3>; rel="next"
    },
 ```
 
-### <a name="get-entries">Get all entries from a register</a>
-
-Endpoint: `GET /entries/`
+### <a name="get-entries">`GET /entries/`</a>
 
 Parameters: 
 * start (Optional): Collection page number. Defaults to 1.
@@ -364,9 +352,7 @@ Link: <?start=101&limit=100>; rel="next"
 ]
 ```
 
-### <a name="get-entries-entry-number">Get a specific entry from a register</a>
-
-Endpoint: `GET /entries/{entry-number}/`
+### <a name="get-entries-entry-number">`GET /entries/{entry-number}/`</a>
 
 ```http
 GET /entries/{entry-number}/ HTTP/1.1
@@ -392,9 +378,7 @@ Content-Type: application/json
 ]
 ```
 
-### <a name="items">Get a specific item within a register</a>
-
-Endpoint: `GET /items/{item-hash}/`
+### <a name="items">`GET /items/{item-hash}/`</a>
 
 ```http
 GET /items/{item-hash}/ HTTP/1.1
@@ -416,9 +400,7 @@ Content-Type: application/json
 }
 ```
 
-### <a name="items">Download the full contents of the register in a ZIP file</a>
-
-Endpoint: `GET /download-register/`
+### <a name="items">`GET /download-register/`</a>
 
 ```http
 GET /download-register/ HTTP/1.1
--- a/source/documentation/using_the_api/api_reference.md
+++ b/source/documentation/using_the_api/api_reference.md
@@ -176,7 +176,7 @@ Content-Type: application/json
 ]
 ```
 
-### <a name="get-records-field-name-field-value">Get all records that share a `{field-value}` for a particular field (`{field-name}`)</a>
+### <a name="get-records-field-name-field-value">Get all records that share a `field-value` for a particular `field-name`</a>
 
 Endpoint: `GET /records/{field-name}/{field-value}/`
 
--- a/source/documentation/using_the_api/api_reference.md
+++ b/source/documentation/using_the_api/api_reference.md
@@ -4,15 +4,15 @@
 
 Endpoint: `GET /register/`
 
-```
-GET /register/ HTTP/2
+```http
+GET /register/ HTTP/1.1
 Host: local-authority-eng.register.gov.uk
 Accept: application/json
 Authorization: YOUR-API-KEY-HERE
 ```
 
-```
-HTTP/2 200
+```http
+HTTP/1.1 200
 Content-Type: application/json
 
 {
@@ -43,20 +43,20 @@ Content-Type: application/json
 Endpoint: `GET /records/`
 
 Parameters: 
-* page-index (Optional): Collection page number. Defaults to 1.
-* page-size (Optional): Collection page size. Defaults to 100. Maximum is 5000. 
+* `page-index` (Optional): Collection page number. Defaults to 1.
+* `page-size` (Optional): Collection page size. Defaults to 100. Maximum is 5000. 
 
-```
-GET /records/ HTTP/2
+```http
+GET /records/?page-index=1&page-size=3 HTTP/1.1
 Host: local-authority-eng.register.gov.uk
 Accept: application/json
 Authorization: YOUR-API-KEY-HERE
 ```
 
-```
-HTTP/2 200
+```http
+HTTP/1.1 200
 Content-Type: application/json
-Link: <?page-index=1&page-size=3>
+Link: <?page-index=2&page-size=3>; rel="next"
 
 {  
    "KIN":{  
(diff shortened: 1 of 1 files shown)


bahmady:
Please note that under `GET /entries/`:

* start (Optional): Collection page number. Defaults to ?.
* limit (Optional): Collection page size. Defaults to ?. Maximum is ?. 

The '?' are placeholders; I don't know what the appropriate values should be so if someone could fill me in that would be 👍

bahmady:
I've tried to remove some of the warnings, since the way we need to do this in the tech docs tool isn't very accessible (see https://github.com/alphagov/tech-docs-template/issues/151 which I opened). 

The pagination warnings I'm hoping are covered by the new format (see for example `GET /records/`) but there are warnings under for example `GET /records/{key}` about having the exact identifier to get a match which this PR removes as things stand. The other one is under `GET /download-register` and warns that it will download _every_ item and entry.

source/documentation/using_the_api/api_reference.md
```diff
@@ -1,22 +1,20 @@
...
-* how many entries and records the register contains
-
-You can check the entry number you have locally is up to date with the register by comparing the value of `total-entries` for each. 
-
-Example URL: `https://local-authority-eng.register.gov.uk/register`
-
-Example request: `curl --request GET --url https://local-authority-eng.register.gov.uk/register/ --header 'Accept: application/json' --header 'Authorization: YOUR-API-KEY-HERE'`
-
-Example response:
+```
```
arnau: If you add the mark `http` any markdown processor like github's or any that relies on highlight.js will highlight this block accordingly.

E.g.

```md
```http
GET /regitser/ HTTP/2
Host: local-authority-eng.register.gov.uk
...
```
```

File: source/documentation/using_the_api/api_reference.md [hunk 2]
```diff
@@ -225,19 +241,26 @@ Example response:
...
 
-### `GET /entries`
+### <a name="get-entries">Get all entries from a register</a>
 
-Get all entries from the register. For example, all updates there have ever been to the `local-authority-eng` register.
+Endpoint: `GET /entries/`
 
-> Results from this API call are paginated. This call will return the first 100 entries from the first page of the register. Use `limit` to define the maximum number of entries you want and `start` to define the entry number you want to start from (in ascending order).
+Parameters: 
+* start (Optional): Collection page number. Defaults to ?.
```
arnau: I presume it defaults to 1. If you do:

```
curl -I -H 'Accept: application/json' 'https://local-authority-eng.register.gov.uk/entries/'
```

You'll see a response such as:

```http
HTTP/1.1 200 OK
Content-Type: application/json
Link: <?start=101&limit=100>; rel="next"
```

So it suggests that to get the next page you need to start at 101.

source/documentation/using_the_api/api_reference.md
```diff
@@ -225,19 +241,26 @@ Example response:
...
+* limit (Optional): Collection page size. Defaults to ?. Maximum is ?.
```
arnau: Based on the above, this defaults to 100. @michaelabenyohai, @gidsg what's the maximum for entries?

source/documentation/using_the_api/api_reference.md
```diff
@@ -225,19 +241,26 @@ Example response:
...
+GET /entries/ HTTP/2
+Host: local-authority-eng.register.gov.uk
+Accept: application/json
+Authorization: YOUR-API-KEY-HERE
+```
 
 ```
+HTTP/2 200
+Content-Type: application/json
+Link: <?start=1&limit=10>
```
arnau: I'm being picky here but `Link` is:

```
Link: <?start=101&limit=100>; rel="next"
```

source/documentation/using_the_api/api_reference.md
```diff
@@ -101,19 +106,21 @@ Example response:
 }
 ```
 
-### `GET /records/{key}`
-
-Get a specific record within a register based on a particular key. For example, the record for the Borough Council of King's Lynn and West Norfolk in the `local-authority-eng` register. 
+### <a name="get-records-key">Get a specific record within a register based on a particular key</a>
 
-> You must have the exact field value of the unique identifier for the record to get a match from the register. For example, in the `local-authority-eng` register, the field value of the unique identifier (the three-letter code for the county council) must be in capital letters.
+Endpoint: `GET /records/{key}/`
```
arnau: To me, `{key}` is a required parameter so it should be listed in a "Parameters" section.

source/documentation/using_the_api/api_reference.md
```diff
@@ -101,19 +106,21 @@ Example response:
...
 
-Example URL: `https://local-authority-eng.register.gov.uk/records/KIN`
-
-Example request: `curl --request GET --url https://local-authority-eng.register.gov.uk/records/KIN --header 'Accept: application/json' --header 'Authorization: YOUR-API-KEY-HERE'`
-
-Example response:
+```
+GET /records/{key} HTTP/2
```
arnau: Here `{key}` should be something specific, given the response, I think the value is `KIN`.

source/documentation/using_the_api/api_reference.md
```diff
@@ -132,17 +139,21 @@ Example response:
 }
 ```
 
-### <a name="record-entries">`GET /records/{key}/entries`</a>
-
-Get all entries for a single record.
+### <a name="get-records-key-entries">Get all entries for a single record based on a particular key</a>
 
+Endpoint: `GET /records/{key}/entries/`
```
arnau: As above, `{key}` is a required parameter.

source/documentation/using_the_api/api_reference.md
```diff
@@ -132,17 +139,21 @@ Example response:
...
 
-Example request: `curl --request GET --url https://local-authority-eng.register.gov.uk/records/KIN/entries/ --header 'Accept: application/json' --header 'Authorization: YOUR-API-KEY-HERE'`
-
-Example response:
+```
+GET /records/{key}/entries/ HTTP/2
```
arnau: `{key}` needs to be something concrete.

source/documentation/using_the_api/api_reference.md
```diff
@@ -165,21 +176,26 @@ Example response:
...
 
-You can then download the latest item. For example, entry 265 in the above code snippet. Follow the [guidance for downloading items](#items).
-
-### `GET /records/{field-name}/{field-value}`
-
-Get all records that share a field-value for a particular field. For example, all local authorities marked as county councils from the `local-authority-eng` register. 
+### <a name="get-records-field-name-field-value">Get all records that share a `field-value` for a particular `field-name`</a>
 
-> Results from this API call are paginated. This call will return the first 100 records from the first page of the register. Use `page-size` to define the number of records you want and `page-index` to define the pages you want. The maximum `page-size` is 5000.
+Endpoint: `GET /records/{field-name}/{field-value}/`
```
arnau: `{field-name}` and `{field-value}` are required parameters.

source/documentation/using_the_api/api_reference.md
```diff
@@ -165,21 +176,26 @@ Example response:
...
-Example URL: `https://local-authority-eng.register.gov.uk/records/local-authority-type/CTY`
+Parameters: 
+* page-index (Optional): Collection page number. Defaults to 1.
+* page-size (Optional): Collection page size. Defaults to 100. Maximum is 5000. 
 
-Example request: `curl --request GET --url https://local-authority-eng.register.gov.uk/records/local-authority-type/CTY --header 'Accept: application/json' --header 'Authorization: YOUR-API-KEY-HERE'`
-
-Example response:
+```
+GET /records/{field-name}/{field-value}/ HTTP/2
```
arnau: Concrete values

source/documentation/using_the_api/api_reference.md
```diff
@@ -165,21 +176,26 @@ Example response:
...
+Host: local-authority-eng.register.gov.uk
+Accept: application/json
+Authorization: YOUR-API-KEY-HERE
+```
 
 ```
+HTTP/2 200
+Content-Type: application/json
+Link: <?page-index=1&page-size=3>
```
arnau: This `Link` header needs probably a `rel=next` and then values should be higher. Also the `GET` part should contain the parameters, at least the `page-size` as it doesn't default to 100.

arnau (changes requested):
I really like the way it's going 👍

source/documentation/using_the_api/api_reference.md
```diff
@@ -1,22 +1,20 @@
...
-
-You can check the entry number you have locally is up to date with the register by comparing the value of `total-entries` for each. 
-
-Example URL: `https://local-authority-eng.register.gov.uk/register`
-
-Example request: `curl --request GET --url https://local-authority-eng.register.gov.uk/register/ --header 'Accept: application/json' --header 'Authorization: YOUR-API-KEY-HERE'`
-
-Example response:
+```
+GET /register/ HTTP/2
```
arnau: After confirming with @michaelabenyohai, all `HTTP/2` need to change to `HTTP/1.1`.

arnau (approved):
Approved but waiting on the PLACEHOLDER max. @michaelabenyohai do you know the answer?

source/documentation/using_the_api/api_reference.md
```diff
@@ -225,19 +249,27 @@ Example response:
...
-### `GET /entries`
+### <a name="get-entries">Get all entries from a register</a>
 
-Get all entries from the register. For example, all updates there have ever been to the `local-authority-eng` register.
+Endpoint: `GET /entries/`
 
-> Results from this API call are paginated. This call will return the first 100 entries from the first page of the register. Use `limit` to define the maximum number of entries you want and `start` to define the entry number you want to start from (in ascending order).
+Parameters: 
+* start (Optional): Collection page number. Defaults to 1.
+* limit (Optional): Collection page size. Defaults to 100. Maximum is PLACEHOLDER.
```
michaelabenyohai: PLACEHOLDER is 5000

bahmady:
Thank you @michaelabenyohai 

If possible I would like to aim for more than just @arnau's approval of this so will leave it up for review for a bit to see if that's achievable

For the record I'm planning a round of user research on the docs (to do myself) which would include this

michaelabenyohai:
It looks good. My only thought is that we have lost the structured nature to the contents on the left hand side. I had thought we got good feedback to the contents listing "GET /register", "GET /records" etc. The headings do make sense, but are they as easy for someone to navigate quickly?

bahmady:
Yes I thought that over the weekend. The best thing to do might be to partly revert and keep the existing (as in, currently live) headers with just the endpoints and then do the new research from there?

michaelabenyohai:
Yeah that might be good. Does that mean we'll lose the descriptions in those headers? Or will you out them elsewhere?

It might be just me that thinks this, but I do think it was easier to skim the contents previously. It was easier to identify what was "API" and what was other helpful stuff.

bahmady:
The descriptions are currently in the 'Quick start guide' under the 'Choose an endpoint' header so they'll still reside there regardless. 

There's no real reason why I can't duplicate the descriptions again under a the endpoint header.

e.g. 

>`GET /something/`
>"a plaintext description"
>
> `HTTP response`

The other point is with the new multipage/dropdown functionality coming in soon the left-hand nav is likely to change again. How exactly it changes will depend on user research outcomes, I would imagine.

arnau:
If we have feedback that having `GET /foo` is better, we should keep it this way.

About the descriptions, I think it's important to have them in context of each endpoint. 

How is the left-hand nav be affected by the multipage functionality? In any case, we will probably not move to multipage during Q2, unless I missed something.

bahmady:
@arnau It depends on what specific parts of the current docs we choose to put into different pages; if we put any in then they will be accessible via a drop-down menu.

bahmady:
That's as things stand anyway... it's STC.

bahmady:
If you guys are OK with it as it is, I'll merge.

arnau:
@bahmady did you decide to get rid of the descriptions?

arnau (approved):
👌

michaelabenyohai (approved):

PR #89, opened 2018-06-29
Fix erroneous slashes

bahmady:
I have to apologise, because I made some mistakes manually inputting these recently.


bahmady:
My fault then, getting confused because for so long we had the form without trailing slashes. I changed them thinking it was correct, then confused myself again :D

PR #93, opened 2018-07-09
Getting Updates content overhaul

subatoi:
This is a first draft of an overhaul of content for the 'Getting updates' section of the API documentation.

### DIFF
--- a/source/documentation/using_the_api/getting_updates.md
+++ b/source/documentation/using_the_api/getting_updates.md
@@ -1,20 +1,97 @@
 ## Getting updates
 
-You can find the latest entry number by looking at the register information (use the [`GET /register` endpoint](#getregister)) and comparing the most recent entry number with your own copy.
+You can use [the `GET /download-register/` endpoint](#download) to download a full new copy of a register.
 
-Example request: `https://local-authority-eng.register.gov.uk/register`
+### See how many updates have been made to a register
 
-Download a full new copy of the register by using [the `GET /download-register` endpoint](#download).
+You can use the [`GET /records/` endpoint](#getrecord) to return all the records from a register. In each record, there is an [`entry-number` property](#entries). The largest value for this is the current size of a register. You can compare this with a local copy. 
 
-Download all updates for one record according to a particular key by [using the `GET /records/{key}/entries` endpoint](#get-record-key-entries):
+Similarly, you can use the [`GET /register/` endpoint](#getregister) to return the `total-entries` property for a register. You can compare this with the same value in a local copy.
 
-Example URL: `https://local-authority-eng.register.gov.uk/records/KIN/entries`
+### Return all entries since your last update
 
-Example request: `curl --request GET --url https://local-authority-eng.register.gov.uk/records/KIN/entries --header 'Accept: application/json' --header 'Authorization: YOUR-API-KEY-HERE'`
+You can use the [`GET /entries/` endpoint](#getentries) with the optional `start` and `limit` parameters to return specific entries to a register. For example, your highest `entry-number` for your copy of the `country` register could be 205, and you may know that there have been 3 updates to the register since you obtained your copy. You could use the following request, setting `start` to `206` and `limit` to `208`:
 
-Example response:
-
+```http
+GET /entries?start=206&limit=208/ HTTP/1.1
+Host: country.register.gov.uk
+Accept: application/json
+Authorization: YOUR-API-KEY-HERE
 ```
+
+```http
+HTTP/1.1 200
+Content-Type: application/json
+Link: <?start=206&limit=208>; rel="next"
+
+[
+  {
+    "index-entry-number": "206",
+    "entry-number": "206",
+    "entry-timestamp": "2017-03-29T14:22:30Z",
+    "key": "GM",
+    "item-hash": [
+      "sha-256:0429375c4fb403288ef816e5dd38a24f192e35b8f55e40cc6266eb25eaef77b1"
+    ]
+  },
+  {
+    "index-entry-number": "207",
+    "entry-number": "207",
+    "entry-timestamp": "2017-10-25T09:52:52Z",
+    "key": "CI",
+    "item-hash": [
+      "sha-256:b3ca21b3b3a795ab9cd1d10f3d447947328406984f8a461b43d9b74b58cccfe8"
+    ]
+  },
+  {
+    "index-entry-number": "208",
+    "entry-number": "208",
+    "entry-timestamp": "2018-06-13T13:54:40Z",
+    "key": "SZ",
+    "item-hash": [
+      "sha-256:f89f36ed8b2a1417237a8e95b810e8ab4ead844277ad7bc7794cb5f83732c976"
+    ]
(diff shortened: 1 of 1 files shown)


bahmady:
This is a first draft of an overhaul of content for the 'Getting updates' section of the API documentation. Any and all feedback welcomed, including if I need to go back to the drawing board :)

source/documentation/using_the_api/getting_updates.md
```diff
@@ -1,20 +1,97 @@
...
+    "item-hash": [
+      "sha-256:f89f36ed8b2a1417237a8e95b810e8ab4ead844277ad7bc7794cb5f83732c976"
+    ]
+  }
+]
+```
+
+To see what the actual update is for each of these entries, you can use the [`GET /items/{item-hash}` endpoint](#items) on the value in the `item-hash` property. 
+
+For example, you could use the following request for entry 208:
```
bahmady: Do we need to say that users must download updates in order?

source/documentation/using_the_api/getting_updates.md
```diff
@@ -1,20 +1,97 @@
 ## Getting updates
 
-You can find the latest entry number by looking at the register information (use the [`GET /register` endpoint](#getregister)) and comparing the most recent entry number with your own copy.
+You can use [the `GET /download-register/` endpoint](#download) to download a full new copy of a register.
```
arnau: This endpoint as the first option makes me grinch. Has anything changed that makes us want to offer this option as the most important way to get updates?

source/documentation/using_the_api/getting_updates.md
```diff
@@ -1,20 +1,97 @@
...
 
-Example request: `https://local-authority-eng.register.gov.uk/register`
+### See how many updates have been made to a register
```
arnau: What do you mean by "how many updates"?

source/documentation/using_the_api/getting_updates.md
```diff
@@ -1,20 +1,97 @@
...
 
-Download a full new copy of the register by using [the `GET /download-register` endpoint](#download).
+You can use the [`GET /records/` endpoint](#getrecord) to return all the records from a register. In each record, there is an [`entry-number` property](#entries). The largest value for this is the current size of a register. You can compare this with a local copy.
```
arnau: I think the preferred way to know the latest (biggest) entry number is via `GET /register/` so probably I would flip this paragraph with the next one and arguably I would remove the suggestion about finding the largest value. It can be an expensive operation for large registers.

Regarding comparing with the local copy, could you elaborate a bit more what is the user expected to do and to get out of it?

source/documentation/using_the_api/getting_updates.md
```diff
@@ -1,20 +1,97 @@
...
+You can use [the `GET /download-register/` endpoint](#download) to download a full new copy of a register.
 
-Example request: `https://local-authority-eng.register.gov.uk/register`
+### See how many updates have been made to a register
 
-Download a full new copy of the register by using [the `GET /download-register` endpoint](#download).
+You can use the [`GET /records/` endpoint](#getrecord) to return all the records from a register. In each record, there is an [`entry-number` property](#entries). The largest value for this is the current size of a register. You can compare this with a local copy. 
 
-Download all updates for one record according to a particular key by [using the `GET /records/{key}/entries` endpoint](#get-record-key-entries):
+Similarly, you can use the [`GET /register/` endpoint](#getregister) to return the `total-entries` property for a register. You can compare this with the same value in a local copy.
```
arnau: The sentence "compare with…" suggests I already have this value stored and it is expected from me to have done this already. I don't know if there is a better way to phrase that this value is useful if you _know_ the previous highest entry number.

source/documentation/using_the_api/getting_updates.md
arnau: Another important bit is: if you got into the trouble to get all records, there is no real need to compare against anything; a full replace will suffice.

source/documentation/using_the_api/getting_updates.md
```diff
@@ -1,20 +1,97 @@
...
 
-Example URL: `https://local-authority-eng.register.gov.uk/records/KIN/entries`
+### Return all entries since your last update
 
-Example request: `curl --request GET --url https://local-authority-eng.register.gov.uk/records/KIN/entries --header 'Accept: application/json' --header 'Authorization: YOUR-API-KEY-HERE'`
+You can use the [`GET /entries/` endpoint](#getentries) with the optional `start` and `limit` parameters to return specific entries to a register. For example, your highest `entry-number` for your copy of the `country` register could be 205, and you may know that there have been 3 updates to the register since you obtained your copy. You could use the following request, setting `start` to `206` and `limit` to `208`:
```
arnau: I have the feeling there are quite a few assumptions mixed with the example and with the generic approach. I'm not sure I can explain it better, let me try.

You start by saying that I can use an endpoint and it has two optional parameters. Here I'm not entirely sure it is relevant that they are optional but anyway.

There is again the idea of knowing the highest entry number in my local copy which suggests to me that it should be one of the first recommendations in the "getting updates" section or at least the section that is about getting _just_ the updates not the full set of records.

In the example you say "[...] you may know that there have been 3 updates [...]". To my understanding, the point of using the `GET /entries/` endpoint is to not have to know in advance how many updates have happened. The example could be just about using the `start` parameter with the entry number you (should) know in advance. The result will be all new entries.

source/documentation/using_the_api/getting_updates.md
```diff
@@ -1,20 +1,97 @@
...
+    "entry-timestamp": "2018-06-13T13:54:40Z",
+    "key": "SZ",
+    "item-hash": [
+      "sha-256:f89f36ed8b2a1417237a8e95b810e8ab4ead844277ad7bc7794cb5f83732c976"
+    ]
+  }
+]
+```
+
+To see what the actual update is for each of these entries, you can use the [`GET /items/{item-hash}` endpoint](#items) on the value in the `item-hash` property.
```
arnau: I would suggest that instead of "the actual update" we use something more on the lines of "the data for the update" or something on these lines. I don't have a solid argument why I think it sounds better, sorry about that.

source/documentation/using_the_api/getting_updates.md
arnau: No. There is no requirement of getting _items_ (not updates) in order. You could not even get them at all, or just a fraction. A situation where a fraction is good enough is if you take the time to compile the subset of new entries that you get after removing all new entries for duplicated keys. This is an awful sentence, let me give you an example:

1. You have the records result of a register at size 205.
2. `GET /entries/?start=205`.
3. The result has 5 entries with the following (number, key) pairs: (206, Z), (207, B), (208, B), (209, A), (210, Z).
4. You want only the new _records_ (as in the latest data for a given key).
5. You take the unique keys from the result set with the latest entry numbers: (208, B), (209, A), (210, Z).
6. `GET /items/{hash}` for all hash in the set.

arnau (changes requested):

PR #96, opened 2018-07-13
Getting updates content overhaul v2

subatoi:
### Context

Made changes based on @arnau's feedback in the last PR

Removed the reference to the download endpoint; I'm not sure if it's relevant, but can reintroduce per feedback on this one :)

Changes to the 'Getting updates' section only, because the existing content doesn't quite cover all possibilities and has been considered a placeholder

### Guidance to review

Existing content can be viewed live https://docs.registers.service.gov.uk/#getting-updates

### DIFF
--- a/source/documentation/using_the_api/getting_updates.md
+++ b/source/documentation/using_the_api/getting_updates.md
@@ -1,19 +1,23 @@
 ## Getting updates
 
-You can use the [`GET /register/` endpoint](#getregister) to obtain the current highest entry number in a register from the `total-entries` property. If you keep track of this value, you can compare it with a new query to the register to find out if there are any updates.
+You can use the [`GET /register/` endpoint](#getregister) to get the current highest entry number in a register from the `total-entries` property. You can then regularly compare it with new queries to the register to see if there are any updates.
 
 ### Return all entries since your last update
 
 You can use the [`GET /entries/` endpoint](#getentries) with the `start` and `limit` parameters to return specific entries to a register. 
 
-For example, the highest entry number from the last time you got data from the `country` register could have been 205. You could then use the following request to find out how many updates there have been since (if any):
+For example, the last highest entry number you got from the `country` register could be 205. You could then use the following request to find out if there are any new entries:
+
+Request:
 
 ```http
-GET /entries/?start=206/ HTTP/1.1
+GET /entries?start=206/ HTTP/1.1
 Host: country.register.gov.uk
 Accept: application/json
 Authorization: YOUR-API-KEY-HERE
 ```
+
+Response:
 
 ```http
 HTTP/1.1 200
@@ -53,14 +57,18 @@ Link: <?start=206>; rel="next"
 
 To see what data is in the updates for each of these entries, you can use the [`GET /items/{item-hash}` endpoint](#items) on the value in the `item-hash` property. 
 
-Following the example above, you could use the following request for entry 208:
+Following the same example, you could use the following request for entry 208:
+
+Request:
 
 ```http
-GET /items/sha-256:f89f36ed8b2a1417237a8e95b810e8ab4ead844277ad7bc7794cb5f83732c976 HTTP/1.1
+GET /items/sha-256:f89f36ed8b2a1417237a8e95b810e8ab4ead844277ad7bc7794cb5f83732c976
 Host: country.register.gov.uk
 Accept: application/json
 Authorization: YOUR-API-KEY-HERE
 ```
+
+Response:
 
 ```http
 {
@@ -73,7 +81,7 @@ Authorization: YOUR-API-KEY-HERE
 
 ### Return all updates for a particular key 
 
-You can use the [`GET /records/{key}/entries/` endpoint](#get-record-key-entries) to download all updates for a record.
+You can use the [`GET /records/{key}/entries/` endpoint](#get-record-key-entries) to download all updates for a record. 
 
 For example, for the `KIN` key in the `local-authority-register`:
 
--- a/source/documentation/using_the_api/getting_updates.md
+++ b/source/documentation/using_the_api/getting_updates.md
@@ -9,7 +9,7 @@ You can use the [`GET /entries/` endpoint](#getentries) with the `start` and `li
 For example, the highest entry number from the last time you got data from the `country` register could have been 205. You could then use the following request to find out how many updates there have been since (if any):
 
 ```http
-GET /entries?start=206/ HTTP/1.1
+GET /entries/?start=206/ HTTP/1.1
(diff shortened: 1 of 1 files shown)


source/documentation/using_the_api/getting_updates.md
```diff
@@ -1,20 +1,93 @@
...
+  }
+]
+```
+
+To see what data is in the updates for each of these entries, you can use the [`GET /items/{item-hash}` endpoint](#items) on the value in the `item-hash` property. 
+
+Following the example above, you could use the following request for entry 208:
+
+```http
+GET /items/sha-256:f89f36ed8b2a1417237a8e95b810e8ab4ead844277ad7bc7794cb5f83732c976
```
arnau:
```
GET /items/sha-256:f89f36ed8b2a1417237a8e95b810e8ab4ead844277ad7bc7794cb5f83732c976 HTTP/1.1
```

arnau (changes requested):
I like it. I made a few suggestions or minor corrections.

arnau (approved):

bahmady:
@arnau you may want to re-review following the latest commit, if time permits

source/documentation/using_the_api/getting_updates.md
```diff
@@ -1,20 +1,103 @@
...
+Accept: application/json
+Authorization: YOUR-API-KEY-HERE
+```
+
+Example response:
+
+```http
+HTTP/1.1 200
+Content-Type: application/json
+Link: <?start=206>; rel="next"
```
michaelabenyohai: This `Link` header is incorrect. It would actually be:

```
Link <?start=106&limit=100>; rel="previous"
```

source/documentation/using_the_api/getting_updates.md
```diff
@@ -1,20 +1,103 @@
...
 
-Download all updates for one record according to a particular key by [using the `GET /records/{key}/entries` endpoint](#get-records-key-entries):
+For example, the last highest entry number you got from the `country` register could be 205. You could then use the following request to find out if there are any new entries:
 
-Example URL: `https://local-authority-eng.register.gov.uk/records/KIN/entries`
+Example request:
 
-Example request: `curl --request GET --url https://local-authority-eng.register.gov.uk/records/KIN/entries --header 'Accept: application/json' --header 'Authorization: YOUR-API-KEY-HERE'`
+```http
+GET /entries?start=206/ HTTP/1.1
```
michaelabenyohai: Note that `limit` defaults to 100. So if there have been more than 100 entries added to the register they will not get all of them by calling `/entries?start206`. Is it worth adding something to specify this?

The user can look at the `Link` header returned. If there are more entries than the 100 returned then they would see a `Link` header with `rel="next"` returned:

```
<?start=306&limit=100>; rel="next"
```

And keep going until they no longer see a `Link` header with `rel="next"`.

Alternatively the user could use the `/register` endpoint to identify the new latest entry-number and calculate the correct limit. E.g. if the new `total-entries` value was `311` they should use `limit=105` (311-206). This will not work if the limit ends up being more than 5000 however, and they would need to follow the link headers anyway.

cc @arnau

arnau (changes requested):
There are two or three regressions, otherwise it's fine

source/documentation/using_the_api/getting_updates.md
```diff
@@ -1,20 +1,103 @@
...
+  "country": "SZ",
+  "official-name": "Kingdom of Eswatini",
+  "name": "Eswatini",
+  "citizen-names": "Swazi"
+}
 ```
+
+### Return all updates for a particular key 
+
+You can use the [`GET /records/{key}/entries/` endpoint](#get-record-key-entries) to download all updates for a record.
```
michaelabenyohai: Describing these as "updates" is perhaps misleading here. This endpoint returns all entries that exist for that key since the beginning of time. They are therefore not updates in the sense that you could get something you already have, unlike the approach in the section above, where you will always be returned entries you don't already have.

It doesn't help that although this endpoint returns entries in the same way that the `/entries` endpoint does, it does not support pagination. This endpoint also currently has a limit of returning 5000 entries. So if there are more than 5000 entries for a key, this request will fail. It might be worth adding that there is a limit in the docs for now. Then we should fix it and consider adding pagination. 

Thoughts @arnau?

source/documentation/using_the_api/getting_updates.md
```diff
@@ -1,20 +1,103 @@
...
+You can use the [`GET /register/` endpoint](#getregister) to get the current highest entry number in a register from the `total-entries` property. You can then regularly compare it with new queries to the register to see if there are any updates.
 
-Example request: `https://local-authority-eng.register.gov.uk/register`
+### Return all entries since your last update
 
-Download a full new copy of the register by using [the `GET /download-register` endpoint](#get-download-register).
+You can use the [`GET /entries/` endpoint](#getentries) with the `start` and `limit` parameters to return specific entries to a register. 
 
-Download all updates for one record according to a particular key by [using the `GET /records/{key}/entries` endpoint](#get-records-key-entries):
+For example, the last highest entry number you got from the `country` register could be 205. You could then use the following request to find out if there are any new entries:
```
michaelabenyohai: We don't seem to specify anywhere how you know what the latest entry number you got was. This may have been removed from a previous version of this - did we deem it unnecessary?

michaelabenyohai (changes requested):

source/documentation/using_the_api/getting_updates.md
michaelabenyohai: Also, should we use "get" rather than "download"?

arnau:
We removed the assumption that said over an over "you store locally". Intead, the first paragraph explains what you should probably do.

source/documentation/using_the_api/getting_updates.md
bahmady: @arnau commented on this section in https://github.com/alphagov/registers-tech-docs/pull/93/files

☝️ that was the original PR for this content which I scrapped because too many changes were needed. 

> There is again the idea of knowing the highest entry number in my local copy which suggests to me that it should be one of the first recommendations in the "getting updates" section or at least the section that is about getting just the updates not the full set of records.

Originally this line:

> You can then regularly compare it with new queries to the register to see if there are any updates.

was this:

> You can keep track of this value and regularly compare it with new queries to the register to see if there are any updates.

("this value" means `total-entries`)

source/documentation/using_the_api/getting_updates.md
arnau: Mmm, I wonder if we should consider adding pagination to `/records/{key}/entries/`. It is a collection that can grow big after all.

PR #97, opened 2018-07-16
Update content types to be consistent with the implementation

MatMoore:
### Context
When I tried specifying the content types in the documentation with an `Accept` header, `text/tsv` and `text/tff` return 406 responses (406 = "not acceptable").

The supported media types are defined in the implmentation here:
https://github.com/openregister/openregister-java/blob/master/src/main/java/uk/gov/register/views/representations/ExtraMediaType.java

These are the correct values according to
- https://www.w3.org/TR/turtle/#sec-mime
- https://www.iana.org/assignments/media-types/text/tab-separated-values


bahmady:
Thanks @MatMoore 👍

PR #105, opened 2018-07-20
Getting updates content overhaul v3

subatoi:
### Context
Scrapped the previous two attempts at this and tried to incorporate all feedback from both into this PR. Hopefully third time's the charm. 

### Changes proposed in this pull request
Overhaul of the 'Getting updates' section to make appropriate recommendations on what users should do at the current time to achieve this.

PR #106, opened 2018-07-23
Spinoff of 'Getting updates v3' - alternative version 

bahmady:
### Context
@arnau made some recommendations based on not reinforcing the use of the `limit` parameter. This is a suggested alternative to https://github.com/alphagov/registers-tech-docs/pull/105 to get around that. 

- Reorganises accordingly (simplifies and removes some content)

- Also adds some accidentally omitted content

- Uses the `government-organisation` register in the examples instead of the `country` register, since it's updated more often

- Changes example to give both `next` and `previous` in the `Link` header

### DIFF
--- a/source/documentation/using_the_api/getting_updates.md
+++ b/source/documentation/using_the_api/getting_updates.md
@@ -6,10 +6,10 @@ You can regularly compare this value with new queries to a register to see if th
 
 ### Return all new entries since your last update
 
-You can use the [`GET /entries/` endpoint](#getentries) to return specific entries to a register. This takes the parameters `start` and `limit`. 
+You can use the [`GET /entries/` endpoint](#getentries) to return specific entries to a register. 
 
 For example, your last highest entry number from the `government-organisation` register could
-be 750. You could then make this request to find out if there are any new entries:
+be 750. To find out if there are any new entries, you could make this request:
 
 ```http
 GET /entries/?start=751 HTTP/1.1
@@ -18,7 +18,7 @@ Accept: application/json
 Authorization: YOUR-API-KEY-HERE
 ```
 
-If you receive a response, you'll know there are updates.
+If you get a response that's not empty, you'll know there are updates.
 
 You should also look at the `Link` header. The response to the example would contain this header:
 
@@ -28,9 +28,7 @@ Content-Type: application/json
 Link: <?start=850&limit=100>; rel="next",<?start=650&limit=100>; rel="previous"
 ```
 
-Here, there are more than 100 new entries since the last update, and `rel="next"` shows that these have been paginated. 
-
-In this situation, you could set the `limit` parameter higher. You could also make new requests, increasing the value of `start` each time, until a `Link` header is returned without `rel="next"`.
+Here, there are more than 100 new entries since the last update. In this situation, you can follow the links in `rel="next"` until the `Link` header no longer contains one. 
 
 ### See what data is in each entry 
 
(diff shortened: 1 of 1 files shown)


source/documentation/using_the_api/getting_updates.md
```diff
@@ -6,106 +6,38 @@ You can regularly compare this value with new queries to a register to see if th
...
 ```
 
-### If you have more updates than the value of `limit` 
+Here, there are more than 100 new entries since the last update, as shown by `rel="next"`. 
 
-[The `limit` parameter defaults to 100, and its maximum value is
-5000](#get-entries). In the previous example, `limit` was 100. Here, if there
-were more than 100 new entries since your last update, you would not get them
-all. 
+In this situation, you could set the `limit` parameter higher. You could also make new requests, increasing the value of `start` each time, until a `Link` header is not returned with `rel="next"`.
```
arnau: I think this paragraph should remove any mention to the `limit` parameter. The point of using `Link` with `rel="next"` links is to not have to explain how pagination works so this should be more on the lines of:

> In this situation you have to make a new request following the link in `rel="next"`, if the next page has another `rel="next"` you know there are entries you still need to consume, the latest page will not have any `rel="next"`.

source/documentation/using_the_api/getting_updates.md
```diff
@@ -6,106 +6,38 @@ You can regularly compare this value with new queries to a register to see if th
 
 ### Return all new entries since your last update
 
-You can use the [`GET /entries/` endpoint](#getentries) with the `start` and
-`limit` parameters to return specific entries to a register. 
+You can use the [`GET /entries/` endpoint](#getentries) to return specific entries to a register. This takes the parameters `start` and `limit`.
```
arnau: I would only mention the use of `start` when it's needed in order to start fetching entries from a custom starting point. And not mention at all `limit`.

source/documentation/using_the_api/getting_updates.md
```diff
@@ -6,106 +6,38 @@ You can regularly compare this value with new queries to a register to see if th
...
-GET /entries/?start=206 HTTP/1.1
-Host: country.register.gov.uk
+GET /entries/?start=751 HTTP/1.1
+Host: goverment-organisation.register.gov.uk
 Accept: application/json
 Authorization: YOUR-API-KEY-HERE
 ```
 
-Example response:
+If you receive a response, you'll know there are updates.
```
arnau: You will always receive a response, and a 200 one. If there are no new entries you'll get an empty collection:

```
HTTP/1.1 200 OK
Content-Type: application/json
Content-Length: 2
Link: <?start=1990&limit=100>; rel="previous"

[]
```

source/documentation/using_the_api/getting_updates.md
```diff
@@ -6,106 +6,33 @@ You can regularly compare this value with new queries to a register to see if th
...
-
-...
-```
-
-You could then make new requests, increasing the value of `start` each time, until
-you no longer see a `Link` header returned with `rel="next"`.
+### See what data is in each entry 
 
+You can use the [`GET /items/{item-hash}` endpoint](#items) on the value in
+the `item-hash` property
```
arnau: Should this sentence end with a full stop (`.`)?

bahmady:
@michaelabenyohai are you happy to go with this version / do you prefer the other version in https://github.com/alphagov/registers-tech-docs/pull/105 / are both not right?

PR #109, opened 2018-07-26
[DONOTMERGE] adds search

subatoi:
### Context
Adds a search bar to the tech docs

Relevant changes to the config files, the rest should be picked up by the build/deploy

### Guidance to review
Testable at https://registers-docs-testing-2.cloudapps.digital/

### DIFF
--- a/config/tech-docs.yml
+++ b/config/tech-docs.yml
@@ -21,3 +21,6 @@ google_site_verification: "IJ5HOXpZrISM6la-YO_iX0rBUC5YFftpexygcKLsNs4"
 # Multi-page options
 multipage_nav: true
 collapsible_nav: true
+
+# Search
+enable_search: true


bahmady:
Bug: https://github.com/alphagov/tech-docs-template/issues/166

gidsg (approved):
🔍 👍

bahmady:
There's a few issues with the search right now which I've fed back to the developers. One is captured here:

https://github.com/alphagov/tech-docs-template/issues/166

☝️ I've tested a fix for this locally and I'm satisfied it's been solved.

One more serious problem is that (and you can confirm this in the test environment at https://registers-docs-testing-2.cloudapps.digital/#gov-uk-registers-quick-start-guide) currently searching for certain things like 'Get' doesn't produce the intended outcome. The developers at Convivio are currently working on it and it's at least partly because of stop words, see https://github.com/alphagov/tech-docs-gem/pull/38

The test environment currently has a test fix though and whilst searching 'Get' (which was one of the stop words) no longer produces 0 results, it doesn't produce expected results in that it doesn't show any of the endpoints. Same goes for searching for e.g. 'Linked'. 

For the purposes of this PR, I think we should hold off before merging. We could still go with it as it stands: I will leave it up to you guys @gidsg @michaelabenyohai @arnau @MatMoore

bahmady:
The fix to the first issue I mention above is also fixed in https://registers-docs-testing-2.cloudapps.digital/#gov-uk-registers-quick-start-guide

MatMoore:
The results look a bit dodgy to me too. If I copy a description from one of them like "Get information about a register" or "Get all records from a register" that page doesn't show up in the results.

The snippet on the results page sometimes looks strange as well, for example if I search for country it shows me the quick start guide with "… //country … country … //country … //country … //country …" 

Do you know if the queries are currently being sent to GA so we can monitor what users actually search for? We don't need it to be perfect to put something live, but we should be able to get some usage data so we can prioritise improvements.

bahmady:
I don't think we have any GA for the test environment(s) but we do for the live environment, though I guess @michaelabenyohai would need to set it up for this search specifically if we merge this PR?

Having spoken to Convivio it seems like lunr is not returning the API reference pages as results in at least some cases. If you try searching "Linked" though, for example, it does return some 🤔

MatMoore:
Ah this is the PR that will add GA tracking: https://github.com/alphagov/tech-docs-gem/pull/39

so it's not there yet, but we'll have it eventually.

arnau:
Not sure what can be done here but results like this one make me wonder if we would be better with a list of links (no excerpt) or with the first paragraph.

screen shot 2018-07-31 at 09 04 07

arnau:
> alphagov/tech-docs-template#166
>
> ☝️ I've tested a fix for this locally and I'm satisfied it's been solved.

@bahmady it doesn't work for me in this branch. Do I have to do anything in particular?

arnau:
> One, more serious, problem is that (and you can confirm this in the test environment at https://registers-docs-testing-2.cloudapps.digital/) currently searching for certain things like 'Get' doesn't produce the intended outcome.

I'm not too bothered by this. A result list for the search “get” is not that useful. Even a search like “get entries” gives me lots of results for “get” which are not relevant to me.

That said, “get + entries” works as I would expect. I guess I want it all :D so probably I would like “get“ to be relevant only when it is in an AND operation. In any case, I rather not have results for “get”.

arnau:
I guess that the best solution for the tech docs gem would be to allow stop-word customisation so we can tune it based on our needs.

arnau:
Another surprising search is “json”. Do we truly only have two pages with “json” in it?

bahmady:
@arnau we can talk about the first comment you raised separately if that's OK? We should be able to have stop-word customisation, yes. 

re. the "json" search, no; I would think it should show more than that.

The problem re. "Get" is with what user research I was able to do on the Pay docs specifically (i.e. not just user research on the format itself) both participants tried to search "Get" and were unsuccessful. I realise that's a limited sample and for the record neither were developers, but nevertheless it's a little concerning. 

re. the other fix, I really don't know: it's fixed for me locally and on that test version ... what are you searching for/where are you when you initiate the search?

lewisnyman:
It seems that the term 'application/json' returns the results you'd also expect. It's not very clear why, but maybe I can add some additional tweaks here to include 'application/json' when searching for 'json'

I've found a fix for the 'get' issue. I've had to make some modifications to the middleman-search gem, once I can get them merged then you can test it.

Per-docs stop word customisation and other customisation options are possible. I'd rather we start trying to get a good default built into the gem first and then add custom options later.

MatMoore:
I'd also rather we had good defaults in the gem. Any project specific customisations will be a pain to manage over time.

lewisnyman:
I've created an issue in the lunr.js repo to see if I can get any pointers: olivernn/lunr.js#365

MatMoore:
We should try this again with the new version of the gem. There are still issues with it but I think it's working better than we first tested this.

If it's still not good enough to use, I can potentially do some work on it to make title matches appear higher in the ranking.

PR #113, opened 2018-08-08
Change header arrangement

subatoi:
### Context
The header arrangement on the live docs as they stand is a little confusing, and makes it look like all the docs are is a 'quick start guide' until further inspection

Changes to the index file in question 

### Guidance to review
Live docs: https://docs.registers.service.gov.uk/
Test docs where this change is pushed: https://registers-docs-testing-2.cloudapps.digital/#gov-uk-registers-documentation

### DIFF
--- a/source/index.html.md.erb
+++ b/source/index.html.md.erb
@@ -3,7 +3,8 @@ title: GOV.UK Registers quick start guide
 weight: 10
 ---
 
-# GOV.UK Registers<br /> quick start guide
+# GOV.UK Registers<br /> documentation
+# Quick start guide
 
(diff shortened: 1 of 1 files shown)


MatMoore:
This might need some input from @injms as it seems like it causes some problem for small screens:

screen shot 2018-08-08 at 17 16 53

I think we should find another way of solving this because having two headers of the same level next to each misrepresents the structure of the page. The markup implies there's a section there but there's nothing in it. This means we get an extra heading added to the sidebar navigation, and I'd expect it to also interfere with screenreaders and search engines indexing the site.

PR #115, opened 2018-08-09
New landing page

bahmady:
### Context
At the moment we don't have the best route for user journeys into the technical documentation, and some feedback we've received indicates that users aren't sure where the "quick start guide" ends and where the "technical documentation" begins. 

This PR effectively introduces a new landing page for the technical documentation, with links to the other (existing) sections. Due to the nature of the technical documentation tool, this is the most straightforward solution I could devise.

Changes to the relevant files to create a better and clearer user journey for users in terms of where their journey on the docs side begins. Also changes the "GOV.UK Registers quick start guide" to just "Quick start guide". 

### Guidance to review
This is up for review on https://registers-docs-testing-2.cloudapps.digital/

The live docs are at: https://docs.registers.service.gov.uk/

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,6 @@ gem 'wdm', '~> 0.1.0', platforms: [:mswin, :mingw]
 # Windows does not come with time zone data
 gem 'tzinfo-data', platforms: [:mswin, :mingw, :jruby]
 
-gem 'govuk_tech_docs', :git => 'https://github.com/alphagov/tech-docs-gem.git', ref:'de0c518324f01cbb107fcd5dc685a44cdca4ac00'
-gem 'middleman-search', git: 'git://github.com/alphagov/middleman-search.git'
+gem 'govuk_tech_docs', '~> 1.5.0'
 
 gem 'therubyracer'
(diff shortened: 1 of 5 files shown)


bahmady:
@mpwoods new version now live at https://registers-docs-testing-2.cloudapps.digital/

bahmady:
@MatMoore @arnau do you have any thoughts on this proposed change?

arnau:
Fine with me.

bahmady:
Cool, I'll merge, the plan is to test it and see how it goes. Obviously if any problems come up we can revert.

MatMoore:
It still really bugs me that "GOV.UK registers technical documentation" shows up as a section in the navigation. It makes it look like there is an extra page that doesn't actually exist.

navigation sidebar with "GOV.UK registers technical documentation" included

And it doesn't get highlighted when you're on that page, like other pages do:
the quick start section with visual highlighting

I think this is probably a limitation of the template rather than something specific to the registers docs, but can we do anything about it?

bahmady:
Unfortunately I'm not sure if there is anything we can do about it with the way the tool is set up. Without proper testing it's hard to know what to do, but I would say that this is clearer than the previous version we had where it wasn't clear where the 'quick start guide' ended and where the tech docs as a whole began. We spoke with Conor about it and he agreed it was best to have the landing page but unfortunately I don't seem to be able to tag him for a review.

Despite all this I'm more than happy to move to something different if we can come up with something else. 

cc @jonathanglassman in case he has any thoughts.

PR #128, opened 2018-08-28
Format changes

subatoi:
### Context
We will be removing support for certain formats.

Our spec is in JSON and most users of registers use JSON
We know this from Analyse historic access by format and by register

We should streamline the number of formats that we offer on registers

Proposed: JSON, CSV, (later on JSON-LD, CSV-LD)

Remove the others: HTML, YAML, Turtle to minimise complexity in the implementation

Remove formats on offer in quick start section that don't conform to proposed changes. This *does not yet have the banner*: we may want a separate PR for that.

### DIFF
--- a/source/api_reference/index.html.md.erb
+++ b/source/api_reference/index.html.md.erb
@@ -1,13 +1,15 @@
 ---
-title: API reference 
+title: API reference
 weight: 50
 ---
 
-# API reference 
+<%= partial 'documentation/data_format_deprecation_notice.md' %>
 
-Each register has an open, RESTful API you can use to access the data. 
+# API reference
 
-Each register API is read-only. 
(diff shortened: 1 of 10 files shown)


arnau:
> This does not yet have the banner: we may want a separate PR for that.

Please, bundle everything in a single PR so we can deploy the changes in a straightforward manner without having to deal with potential complex merges.

arnau (approved):

arnau (changes requested):
When you navigate from section to section via the menu, the scroll is positioned based on the main header. This means the banner is almost always outside the viewport.

I would suggest to put a) the banner after the main header or b) do some sort of repositioning in CSS.

I suspect (b) is either unfeasible or not worth it. Worth checking with conor

PR #131, opened 2018-09-07
Datatypes

subatoi:
### Context
There should be some average-user-friendly content on datatypes in the tech docs

First attempt at adding some content on datatypes, ported from the spec

### DIFF
--- a/source/documentation/datatypes/datatypes.md
+++ b/source/documentation/datatypes/datatypes.md
@@ -1,6 +1,10 @@
+## Boolean
+
+A datatype which can have one of two states, `True` and `False`. 
+
 ## CURIE
 
-Inspired by the [CURIE Syntax 1.0](https://www.w3.org/TR/curie/), the CURIE
+Inspired by the [CURIE Syntax 1.0](https://www.w3.org/TR/curie/), the `CURIE`
 (compact URL) datatype provides a mechanism [to link
 registers](/linked_registers).
 
@@ -10,7 +14,9 @@ URL `https://{name-of-register}.register.gov.uk/records/`.
 ## Datetime
 
 A datatype for date- and time-related data which conforms to a subset of
-[ISO8601](https://www.iso.org/standard/40874.html). The date component is always required.
+[ISO8601](https://www.iso.org/standard/40874.html). 
+
+The date component is always required.
 
 The following are valid `Datetime` values:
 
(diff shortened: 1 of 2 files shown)


source/documentation/datatypes/datatypes.md
```diff
@@ -0,0 +1,85 @@
+## Boolean
+
+A datatype which can have one of two states, `True` and `False`.
```
arnau: I think datatypes in this documentation have to explain how are they returned in the API. And that is as strings so, I would suggest you do an into explaining that and then for each datatype explain what it is and what does it look like in its string form.

E.g. A boolean in its string form is either "true" or "false".

source/documentation/datatypes/datatypes.md
```diff
@@ -0,0 +1,85 @@
...
+A datatype which can have one of two states, `True` and `False`. 
+
+## CURIE
+
+Inspired by the [CURIE Syntax 1.0](https://www.w3.org/TR/curie/), the `CURIE`
+(compact URL) datatype provides a mechanism [to link
+registers](/linked_registers).
+
+CURIEs are of the form `{prefix}:{key}`, where `{prefix}` is mapped to a base
+URL `https://{name-of-register}.register.gov.uk/records/`.
```
arnau: In here you are relying on the string representation but you don't let users know that.

source/documentation/datatypes/datatypes.md
```diff
@@ -0,0 +1,85 @@
...
+```
+2007-03-01T13:00:00Z/2008-05-11T15:30:00Z
+2007-03-01T13:00:00Z/P1Y2M10DT2H30M
+P1Y2M10DT2H30M/2008-05-11T15:30:00Z
+P1Y2M10DT2H30M
+```
+
+## String
+
+A datatype for sequences of one or more Unicode characters.
```
arnau: At this level, a user only needs to know that they should expect a UTF-8 string. And at this level empty strings are not invalid, they mean _null value_.

source/documentation/datatypes/datatypes.md
```diff
@@ -0,0 +1,85 @@
...
+
+Must be encoded in UTF-8.
+
+Empty strings are invalid.
+
+## Text
+
+A subset of the [`String`](#string) datatype. For sequences of Unicode
+characters formatted in Markdown.
```
arnau: Don't go into "sequences of Unicode", users of this documentation are not necessarily familiar with Unicode.

source/documentation/datatypes/datatypes.md
```diff
@@ -0,0 +1,85 @@
...
+
+The following is a valid `Timestamp` value:
+
+```
+2018-07-15T14:38:05Z
+```
+
+## URL
+
+A datatype to contain an absolute URL.
```
arnau: Call me picky but the datatype does not contain anything, it _is_ an absolute URL. If you want, you can say that it _represents_ an absolute URL.

arnau (changes requested):
I think it needs a revisit to set the right expectations. The user of this documentation knows about REST APIs, JSON, CSV and some programming languages but doesn't necessarily know low-level stuff.

source/documentation/datatypes/datatypes.md
bahmady: > And at this level empty strings are not invalid, they mean null value.

Is it worth mentioning, then?

arnau: I think it is important to explain what is the meaning of an empty string, yes. Probably outside the "String datatype" section because it applies to any datatype.

PR #141, opened 2018-11-09
Add a version history to the API documentation

MatMoore:
### Context
We only have 1 version at the moment, but are working on version 2.

- Explain how to request a different version of the API
- Add a changelog to the API reference

This should mirror the specification changelog (https://spec.openregister.org/v2/changelog) except that
- It will only contain the changes we have implemented
- It may include additional changes we've made that are outside of the specification or specific to the way we've chosen to implement something
- The audience is API users, not implementors

### Guidance to review
We are working on additional guidance around our versioning strategy and I expect us to do more work on the docs before we actually release v2.

In particular, there is currently no way to select which version to show in the API reference, but documenting the changes as we go will make this easier to add later.

Related trello cards:
- https://trello.com/c/VUhFKnKy/3033-implement-rfc0024-indexes-removal-breaking-changes
- https://trello.com/c/gRvrLxdY/2987-implement-rfc0016-blob-resource


source/api_reference/version-history.html.md
```diff
@@ -0,0 +1,30 @@
...
+title: Version history
+weight: 0
+---
+# Version history
+
+## Version 1
+
+Version 1 can be accessed using either of these base URLs:
+
+- `https://{register-name}.register.gov.uk/v1/`
```
arnau: We could suggest `/v1` is the preferred way (I imagine the other one will end up being a redirection)

source/quick_start_guide/index.html.md.erb
```diff
@@ -25,11 +25,20 @@ The following are some examples of base URLs:
 
 <div style="height:1px;font-size:1px;">&nbsp;</div>
 
-### Choose an endpoint  
+### Append the current version of the API
+The current version of the API is **v1**.
+
+You should add the version of the API you want to use to the end of the base URL. This step is optional for version 1 of the API and will become required when version 1 is no longer supported.
+
+For example, the full Base URL for version 1 of the `local-authority-eng` register is `https://local-authority-eng.register.gov.uk/v1/`.
```
arnau: Is it a expected that `https://local-authority-eng.register.gov.uk/v1` returns a 404?

source/api_reference/version-history.html.md
MatMoore: It's implemented so that /v1/ is the redirect. I don't think it matters much though, and the plan is to remove it in a few months anyway.

source/quick_start_guide/index.html.md.erb
MatMoore: Nope, that's not ideal. I'll raise a PR to fix that.

source/api_reference/index.html.md.erb
```diff
@@ -27,6 +27,5 @@ Using different endpoints, you can:
 * get a [specific item within a register](/api_reference/get_items_item_hash#get-items-item-hash)
 * download the [full contents of a register in a ZIP file](/api_reference/get_download_register#get-download-register)
 
-You can also read about [HTTP status codes](/api_reference/http_status_codes)
-and [rate limits](/api_reference/rate_limits).
+You can also read about [HTTP status codes](/api_reference/http_status_codes), [rate limits](/api_reference/rate_limits) and [version numbers](/api_reference/version-history.html).
```
bahmady: It might be a good idea to bullet these three items: it'll be easer and more accessible for users. 

Also, I'm not sure the `.html.md` extension is needed for the `version-history` file unless I'm missing some utility of it. It makes linking a little bit more complicated, as here.

source/api_reference/version-history.html.md
```diff
@@ -0,0 +1,31 @@
+---
+title: Version history
+weight: 0
+---
+# Version history
+
+## Version 1
+
+Version 1 can be accessed using either of these base URLs:
```
bahmady: Perhaps:

> You can access Version 1 using these base URLs:

source/api_reference/version-history.html.md
```diff
@@ -0,0 +1,31 @@
...
+
+- `https://{register-name}.register.gov.uk/v1/`
+- `https://{register-name}.register.gov.uk/`
+
+## Next version
```
bahmady: For style consistency it would be good to have spaces between headers and content.

source/api_reference/version-history.html.md
```diff
@@ -0,0 +1,31 @@
...
+You can use this version of the API to test against backwards incompatable API changes before we release a new version of the API.
+
+Access it using this base URL:
+
+- `https://{register-name}.register.gov.uk/next/`
+
+**WARNING: Do not use this URL in production, as its API may change without warning.**
```
bahmady: Unfortunately we shouldn't use bold for anything other than interface text. You could use the solution Jon Heslop came up with for me on Pay to get the warning symbol out of the Design System: https://github.com/alphagov/pay-tech-docs/blob/master/source/payment_flow_overview/index.html.md.erb

source/api_reference/version-history.html.md
```diff
@@ -0,0 +1,31 @@
...
+title: Version history
+weight: 0
+---
+# Version history
+
+## Version 1
+
+Version 1 can be accessed using either of these base URLs:
+
+- `https://{register-name}.register.gov.uk/v1/`
```
bahmady: On https://docs.registers.service.gov.uk/linked_registers/ I've used the placeholder `{register}` rather than `{register-name}`: I would recommend changing this to match that or that to match this for consistency's sake.

source/api_reference/version-history.html.md
bahmady: Also might "v1" be better than "Version 1"?

source/api_reference/version-history.html.md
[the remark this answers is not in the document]
bahmady: It might also be good to introduce a short preamble describing exactly what the 'Next' API is: i.e. it's a term used for a 'preview' of the upcoming API, prior to that API becoming stable.

source/api_reference/version-history.html.md
```diff
@@ -0,0 +1,31 @@
...
+- `https://{register-name}.register.gov.uk/`
+
+## Next version
+You can use this version of the API to test against backwards incompatable API changes before we release a new version of the API.
```
bahmady: Re. the term 'backwards incompatible', I assume users will know what such changes will be before they come to actually test this version of the API, so perhaps:

> You can use the 'Next' version of the API to test against backwards-incompatible changes before 'Next' becomes the stable API. 

I don't know if the 'Next' API will replace the older API entirely or whether the older API will still be hosted? The sentence I've recommended might sound slightly awkward but conventionally there's no "we/us" in the Registers docs at the moment.

source/api_reference/version-history.html.md
```diff
@@ -0,0 +1,31 @@
...
+
+Access it using this base URL:
+
+- `https://{register-name}.register.gov.uk/next/`
+
+**WARNING: Do not use this URL in production, as its API may change without warning.**
+
+### Changes since version 1
```
bahmady: I'm not sure about "Changes since Version 1" if 'Next' API is to be a rolling term. It also may end up being hard to maintain this and this section may slip into obsolescence if nobody tends to it regularly.

source/api_reference/version-history.html.md
```diff
@@ -0,0 +1,31 @@
...
+
+#### Data model changes
+The term "blob" replaces what was previously called an "item" ([RFC-14](https://github.com/openregister/registers-rfcs/blob/master/content/blob-resource/index.md)).
```
bahmady: Perhaps:

> The term "blob" replaces the term "item" ([RFC-14](https://github.com/openregister/registers-rfcs/blob/master/content/blob-resource/index.md)).

source/api_reference/version-history.html.md
```diff
@@ -0,0 +1,31 @@
...
+
+This means that:
```
bahmady: Perhaps:

> This means that:
>
>* you now access the resource at `/blobs/{blob-hash}` instead of `/items/{item-hash}`
>* the field `item-hash` is now named `blob-hash`

source/quick_start_guide/index.html.md.erb
bahmady: No need to capitalise "Base URL" here.

source/quick_start_guide/index.html.md.erb
```diff
@@ -25,11 +25,20 @@ The following are some examples of base URLs:
 
 <div style="height:1px;font-size:1px;">&nbsp;</div>
 
-### Choose an endpoint  
+### Append the current version of the API
+The current version of the API is **v1**.
+
+You should add the version of the API you want to use to the end of the base URL. This step is optional for version 1 of the API and will become required when version 1 is no longer supported.
```
bahmady: Again not sure if we should use the term "v1" throughout rather than "version 1" or "Version 1".

Perhaps:

> ... This is optional for v1 of the API. When v1 is no longer supported, this will be required.

source/api_reference/version-history.html.md
MatMoore: That wording is better, although I think it should say something like "before 'next' is renamed to the next stable API version" rather than just "becomes the stable API"

The timeline will be:
- we make changes to the 'next' version
- when we're happy with it, 'next' becomes version 2
- after a grace period, we get rid of version 1

source/quick_start_guide/index.html.md.erb
MatMoore: that makes sense - I think it's better to be consistent with what's in the URL

source/api_reference/version-history.html.md
bahmady: You could change the content up a bit and describe the timeline itself. For example:

> The release schedule is:
> 
> 1. The GOV.UK Registers team releases the 'Next' API in beta. You can use the 'Next' version of the API to test against backwards-incompatible changes before 'Next' becomes the stable API.
> 2. When the beta is concluded, the 'Next API' becomes the stable API (eg v2). 
> 3. After a set amount of time, the previous version of the API is deprecated.

source/api_reference/version-history.html.md
MatMoore: I was intending this to work like the "unreleased" sections in our changelogs, so that we can document the next version straight away, and then we just rename the heading when `/next` becomes `/v2` or `/v3`.

I'll have a think about how to make this easier to maintain. There are various other things we need to remember to do when making breaking changes as well as this, so it would be great to have a checklist or something.

MatMoore:
Thanks for the review @bahmady! This is all very helpful feedback.

source/api_reference/version-history.html.md
bahmady: There's nothing wrong with it in principle; it's just in my experience of working on different projects, "Version history" sections sometimes end up being very outdated.

bahmady:
This seems to build strangely for me:

image

MatMoore:
@bahmady whoops, I'll add in that banner.

That changes section is meant to be part of the "next" section. I.e. these are the changes that went into that version. v1 would ideally have this too, except for v1 it's just everything.

Can you think of a better way to structure this? I can potentially flatten it, or I can reduce the level of subheadings shown in the navigation.

There will indeed be more changes besides the data model ones, I've just added what's implemented already.

My intention was to use `{register}` and style `v1` and `v2` as code throughout - I'll go through and fix the ones that are inconsistent with that.

bahmady:
@MatMoore 👍 

> That changes section is meant to be part of the "next" section. I.e. these are the changes that went into that version. v1 would ideally have this too, except for v1 it's just everything.

OK, that makes sense. 

> Can you think of a better way to structure this? I can potentially flatten it, or I can reduce the level of subheadings shown in the navigation.

You could just use an h4 for the "Data model changes": h4s shouldn't show up in the left-hand nav AFAICR. 

> My intention was to use {register} and style v1 and v2 as code throughout - I'll go through and fix the ones that are inconsistent with that.

Cool, and that makes sense: it's also consistent with elsewhere.

bahmady:
The `max_toc_heading_level` seems to need to be `2` to have the intended effect but otherwise 👍

bahmady:
Also I've just realised I still can't see the banner

MatMoore:
Have added in the banner now. The template seems to give strange results when the heading level is 2 (it hides so I'm going to leave it at 3 for now.

I think it's because a bunch of pages don't have level 2 headings (it jumps from 1 to 3) so the collapsable menu entries end up empty.

I think this is actually something we should improve in the gem, because at the moment it's purely defined in terms of the html heading levels, rather than how many levels it would actually create in the nav bar. I noticed this also gives unexpected results if you nest directories, because that creates additional nesting that doesn't count towards the limit.

bahmady:
I agree: I think it's more a problem with the tool itself rather than this content.

PR #142, opened 2018-11-20
Updates link to registers collection

injms:
### Context
The navigation link to the registers has been changed to read 'Registers collection' on the frontend; this change does the same for the documentation navigation. (See this [card](https://trello.com/c/GDGJm4ql/3067-update-header-and-footer))

 * Link text changed from 'Get data' to 'Registers collection'

### Guidance to review
None.


injms:
Thanks for approving so quickly @bahmady!

PR #146, opened 2018-12-11
Document new /blobs endpoint

MatMoore:
### Context
`/blobs` is implemented now (I'll deploy it tomorrow)

I've added a note about the new endpoint and how to use it.

I've also restructured this page to work better with the navigation sidebar.

The intention is that when we release this new API version, we will rename this thing from "Upgrading from v1 to next" to "Upgrading from v1 to v2", and add sections as needed when creating new versions.

At some point we will also need to create a page for `GET /blobs` but in order to do this we need a way of structuring content by version number. How we do this depends on whether we accept
https://github.com/openregister/registers-rfcs/pull/43 and just autogenerate the reference documentation. For now I'm just focusing on the "what's changed" aspect.

### DIFF
--- a/source/api_reference/version-history/index.html.md.erb
+++ b/source/api_reference/version-history/index.html.md.erb
@@ -46,9 +46,13 @@ These changes mean that:
 - `/records` now returns an array of record objects, instead of an object indexed by entry key
 - `/records/{field-name}/{field-value}` has moved to `name` and `value` query parameters on `/records`, for example `/records/local-authority-type/CTY` has moved to `/records?name=local-authority-type&value=CTY`
 
-### Added `/blobs`
-You can now use the `/blobs` endpoint to retrieve multiple blobs in a single request. This is faster than fetching blobs one by one when using the `/entries` endpoint.
+### New `/blobs` endpoint
+You can use the `/blobs` endpoint to retrieve multiple blobs in a single request. This is faster than fetching blobs one by one.
 
 `/blobs` is paginated, and by default it returns the first 100 blobs in the register. To fetch more data, query the URL provided in the `Link` header that has `rel=next`.
 
 Blobs are returned in the order they were added to the register. This guarantees that complete pages will always contain the same blobs.
+
+### Hash values are formatted using multihash
+
+We now use the [multihash](https://multiformats.io/multihash/) format to format hash values. This means that blob hashes are now prefixed with the string `1220` instead of `sha-256:`.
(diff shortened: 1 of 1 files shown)


source/api_reference/version-history/index.html.md.erb
```diff
@@ -39,5 +41,13 @@ These changes mean that:
...
\ No newline at end of file
+- `/records/{key}` now returns a flat record object with the entry key using the reserved field `_id`,  instead of an object indexed by the entry key. In addition the `entry-number` and `entry-timestamp` have been removed from the object. If you require these you can get them using the `/records/{key}/entries` endpoint.
+- `/records` now returns an array of record objects, instead of an object indexed by entry key
+
+### Added `/blobs`
+You can now use the `/blobs` endpoint to retrieve multiple blobs in a single request. This is faster than fetching blobs one by one when using the `/entries` endpoint.
+
+`/blobs` is paginated, and by default it returns the first 100 blobs in the register. To fetch more data, query the URL provided in the `Link` header that has `rel=next`.
+
+If you need to incrementally update a copy of the register, you can store the last URL you queried, and query it again to check for updates. You do not need to fetch earlier pages
```
arnau: Is this last paragraph trying to cover the fact that blobs are ordered and so pages never change? If this is the case I would rather say that instead of providing a solution.

MatMoore: Yes - I will try and reword this. I do want to explain a bit why the endpoint is useful though

source/api_reference/version-history/index.html.md.erb
```diff
@@ -39,6 +41,18 @@ These changes mean that:
...
+
+- `/records/{key}` now returns a flat record object with the entry key using the reserved field `_id`,  instead of an object indexed by the entry key. In addition the `entry-number` and `entry-timestamp` have been removed from the object. If you require these you can get them using the `/records/{key}/entries` endpoint.
 - `/records` now returns an array of record objects, instead of an object indexed by entry key
-- `/records/{field-name}/{field-value}` has moved to `name` and `value` query parameters on `/records`, for example `/records/local-authority-type/CTY` has moved to `/records?name=local-authority-type&value=CTY`
\ No newline at end of file
+- `/records/{field-name}/{field-value}` has moved to `name` and `value` query parameters on `/records`, for example `/records/local-authority-type/CTY` has moved to `/records?name=local-authority-type&value=CTY`
+
+### New `/blobs` endpoint
+You can now use the `/blobs` endpoint to retrieve multiple blobs in a single request. This is faster than fetching blobs one by one when using the `/entries` endpoint.
```
gidsg: I would get rid of the word *now* here.

gidsg: Did you mean when using the `/item` endpoint?

PR #148, opened 2018-12-21
Add changenotes for the context endpoint

MatMoore:
### Context
We replaced `/register` with `/context` in V2.

Explain what's changed

### DIFF
--- a/source/api_reference/version-history/index.html.md.erb
+++ b/source/api_reference/version-history/index.html.md.erb
@@ -37,9 +37,9 @@ The registers data model has changed since v1 of the API:
 These changes mean that:
 
 - you now access the resource at `/blobs/{blob-hash}` instead of `/items/{item-hash}`
-- the field `item-hash` has been replaced by `blob-hash`
+- the `item-hash` attribute has been replaced by `blob-hash`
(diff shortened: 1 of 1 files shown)


source/api_reference/version-history/index.html.md.erb
```diff
@@ -53,6 +53,17 @@ You can use the `/blobs` endpoint to retrieve multiple blobs in a single request
 
 Blobs are returned in the order they were added to the register. This guarantees that complete pages will always contain the same blobs.
 
+### New `/context` endpoint
+The context endpoint replaces `/register` ([RFC 18](https://github.com/openregister/registers-rfcs/blob/master/content/context-resource/index.md)).
+
+This endpoint exposes extra metadata about the schema of a register. Each register field is now an object containing `id`, `datatype` and `cardinality` attributes.
```
bahmady: I'm just slightly confused about "field" in the context of this sentence:

> Each register field is now an object containing `id`, `datatype` and `cardinality` attributes.

Since it seems like this PR is pushing for use of the term "attribute". I'm probably missing something so don't want to hold up merging.

